PR #2, opened 2016-03-26
Locking optimization / performance improvement

vladmovchan:
Replace multiple cases of suboptimal use of locking (mutexes + sleep loops) with POSIX semaphores.
This change (mostly commits 9343666 and 6055fa0) significantly reduce load caused by spine process to operating system.
Some comparison numbers available in commit message of 9343666.

### DIFF
--- a/common.h
+++ b/common.h
@@ -87,6 +87,7 @@
 #include <math.h>
 #include <mysql.h>
 #include <netdb.h>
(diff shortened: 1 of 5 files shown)


cigamit:
great work.  will be merging shortly.

vladmovchan:
Thank you!

PR #5, opened 2017-02-03
spine, not spine.c

jer-gentoo:
With enough make jobs, help2man will fail if help2man is run before spine has been built.


ronytomen:
Good catch, I always run it with only -j1

PR #96, opened 2019-06-12
Update configure.ac

whsir:
version number


netniV:
Thanks for the notice and pull request but for now I will leave updating the version number to the next reason.

Issue #208, opened 2021-07-06
segmentation fault in spine v 1.2.18


majed17:
running from command line gives:
```
sudo -u apache ./spine
SPINE: Using spine config file [/etc/spine.conf]
Version 1.2.18 starting
2021/07/06 20:28:55 - SPINE: Poller[1] PID[579] ERROR: Spine Timed Out While Processing Available Thread Lock
FATAL: Spine Encountered a Segmentation Fault
Generating backtrace...0 line(s)...
```

log:

```
2021/07/06 20:30:05 - SPINE: Poller[Main Poller] PID[693] ERROR: Spine Timed Out While Processing Available Thread Lock
--
2021/07/06 20:30:05 - SPINE: Poller[Main Poller] PID[681] ERROR: Spine Timed Out While Processing Available Thread Lock
2021/07/06 20:30:05 - SPINE: Poller[Main Poller] PID[669] ERROR: Spine Timed Out While Processing Available Thread Lock
```

when it was graphing but with gaps, log displayed:

```
2021/07/06 17:45:02 - POLLER: Poller[Main Poller] PID[549085] WARNING: There are 1 processes detected as overrunning a polling cycle, please investigate
--
2021/07/06 17:40:10 - SYSTEM STATS: Time:8.7933 Method:spine  Processes:4 Threads:4 Hosts:39 HostsPerProcess:10 DataSources:4904  RRDsProcessed:1595
2021/07/06 17:40:09 - SPINE: Poller[Main Poller] PID[520420] ERROR: Spine Timed Out While Processing Available Thread Lock
 
 
 
 
2021/07/06 17:40:03 - SPINE: Poller[Main Poller] PID[519954] ERROR: Spine Timed Out While Processing Available Thread Lock
2021/07/06 17:35:05 - SYSTEM STATS: Time:3.7794 Method:spine  Processes:4 Threads:4 Hosts:39 HostsPerProcess:10 DataSources:4904  RRDsProcessed:1299
2021/07/06 17:35:03 - SPINE: Poller[Main Poller] PID[505893] ERROR: Spine Timed Out While Processing Available Thread Lock
2021/07/06 17:35:03 - SPINE: Poller[Main Poller] PID[505921] ERROR: Spine Timed Out While Processing Available Thread Lock
2021/07/06 17:35:01 - SYSTEM WARNING: Primary Admin account notifications disabled!  Unable to send administrative Email.
[...]

PR #215, opened 2021-07-07
Use sem_wait() instead of sem_trywait()

eriksejr:
I think what is going on here is that once spine has spawned the maximum number of threads allowed by the semaphore it only waits 10000 microseconds * 100 retries (1 second) for a thread to complete so it can spawn another. This results in devices/hosts getting skipped and timeout errors appearing in the log if all active threads take longer then 1 second to run. In my environment this happens often and causes spine to not properly poll all hosts. I changed the error handling based on my comments in #208 as the Linux semaphore implementation does not seem to match the way the error handling is done, nor does it seem to ever return some of the errno's that are being checked for in the code. This could be different for a different OS with a different implementation though.

I'd argue that breaking a loop immediately on a successful sem_trywait() is more or less the same thing anyway but concurrency in C is not really my thing. You could also setup a timeout with a signal handler and SIGALRM.

Another option here would be to use sem_timedwait() but this does not really solve the problem, it's just a different way of effectively doing the same thing.

I don't think you'd want to merge this as is, but it is working for me and I hope it's helpful anyway. The implementation using sem_timedwait() in 1.2.17 does not seem to have this problem, I'm not sure what the reason was behind the change.

### DIFF
--- a/spine.c
+++ b/spine.c
@@ -774,6 +774,7 @@ int main(int argc, char *argv[]) {
 				die("ERROR: Invalid sempaphore in call to sem_wait()");
 			} else if (errno = EINTR) {
 				SPINE_LOG_DEVDBG(("WARNING: Interrupted by signal while processing Available Thread Lock"));
+				break;
 			}
 		}
 
@@ -784,6 +785,7 @@ int main(int argc, char *argv[]) {
                                 die("ERROR: Invalid sempaphore in call to sem_wait()");
                         } else if (errno = EINTR) {
                                 SPINE_LOG_DEVDBG(("WARNING: Interrupted by signal while processing Thread Initialization Lock"));
+				break;
                         }
                 }
 
--- a/spine.c
+++ b/spine.c
@@ -765,60 +765,27 @@ int main(int argc, char *argv[]) {
 		details[device_counter]          = poller_details;
 
 		/* dev note - errno was never primed at this point in previous version of code */
-		int wait_retries = 0;
 		int sem_err = 0;
 
-		while (++wait_retries < 100) {
-			sem_err = sem_trywait(&available_threads);
+		sem_err = sem_wait(&available_threads);
 
-			if (sem_err == 0) {
-				break;
-			} else if (sem_err == EAGAIN || sem_err == EWOULDBLOCK) {
-				SPINE_LOG_DEVDBG(("WARNING: Spine Timed Out While Processing Available Thread Lock Internal"));
-			} else {
-				SPINE_LOG_DEVDBG(("WARNING: Spine Error %ld While Processing Available Thread Lock Internal", sem_err));
+		if (sem_err != 0) {
+			if (errno == EINVAL) {
+				die("ERROR: Invalid sempaphore in call to sem_wait()");
+			} else if (errno = EINTR) {
+				SPINE_LOG_DEVDBG(("WARNING: Interrupted by signal while processing Available Thread Lock"));
 			}
-
-			usleep(10000);
 		}
 
-		if (sem_err == EDEADLK) {
-			SPINE_LOG(("ERROR: Spine would have deadlocked Available Thread Lock"));
-			break;
-		} else if (sem_err == ETIMEDOUT || sem_err == EWOULDBLOCK || wait_retries >= 100) {
-			SPINE_LOG(("ERROR: Spine Timed Out While Processing Available Thread Lock"));
-			break;
-		} else if (sem_err != 0) {
-			SPINE_LOG(("ERROR: Spine Encounter Unexpected Error %d For Available Thread Lock", sem_err));
-			break;
-		}
-
-		wait_retries = 0;
-
-		while (++wait_retries < 100) {
-			sem_err = sem_trywait(&thread_init_sem);
-
-			if (sem_err == 0) {
-				break;
-			} else if (sem_err == EAGAIN || sem_err == EWOULDBLOCK) {
-				SPINE_LOG_DEVDBG(("WARNING: Spine Timed Out While Processing Thread Initialization Lock"));
-			} else {
-				SPINE_LOG_DEVDBG(("WARNING: Spine Error %ld While Processing Thread Initialization Lock", sem_err));
-			}
-
-			usleep(10000);
-		}
-
-		if (sem_err == EDEADLK) {
-			SPINE_LOG(("ERROR: Spine would have deadlocked Thread Initialization Lock"));
-			break;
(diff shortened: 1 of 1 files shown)


majed17:
```
sudo make
gcc -DHAVE_CONFIG_H -I. -I./config     -I/usr/include/net-snmp -I/usr/include/net-snmp/.. -I/usr/include/mysql -g -O2 -MT spine.o -MD -MP -MF .deps/spine.Tpo -c -o spine.o spine.c
mv -f .deps/spine.Tpo .deps/spine.Po
/bin/sh ./libtool  --tag=CC   --mode=link gcc  -I/usr/include/net-snmp -I/usr/include/net-snmp/.. -I/usr/include/mysql -g -O2  -L/usr/lib64 -L/usr/lib64  -o spine sql.o spine.o util.o snmp.o locks.o poller.o nft_popen.o php.o ping.o keywords.o error.o  -lnetsnmp -lmysqlclient -lm -ldl -lmysqlclient -lm -ldl -lcrypto -lz -lpthread -ldl -lm -lpthread -lssl 
libtool: link: gcc -I/usr/include/net-snmp -I/usr/include/net-snmp/.. -I/usr/include/mysql -g -O2 -o spine sql.o spine.o util.o snmp.o locks.o poller.o nft_popen.o php.o ping.o keywords.o error.o  -L/usr/lib64 -lnetsnmp -lmysqlclient -lcrypto -lz -ldl -lm -lpthread -lssl
/usr/bin/help2man --output=spine.1 --name='Data Collector for Cacti' --no-info --version-option='--version' ./spine

```
```
sudo -u apache ./spine
SPINE: Using spine config file [/etc/spine.conf]
Version 1.2.18 starting
Time: 8.7277 s, Threads: 4, Devices: 39

```
clean output

eriksejr:
Add link to issue #208

netniV:
Thanks for the suggestion, but the whole reason to use the trywait is for reporting and pretty much reverts that change (which did previously use sem_wait).  If a longer period is needed, then we should extend that period rather than switching back to a blocking wait.

eriksejr:
No problem. If you do go back to the other way with a longer timeout you may wish to remove until_spec from the code since it is not used. If you revert to 1.2.17's way I noticed a bit of an issue here with the code while reading over it:

```
/* specify the point of timeout for timedwait semaphores */
        until_spec.tv_sec = (time_t)(set.poller_interval + begin_time - 0.2);
        until_spec.tv_nsec = 0;
```
The result of the calculation for tv_sec is going to be a float but on most systems time_t is going t be a signed int (32 or 64 bits is platform dependant). If that is type-casted that to time_t I don't think you are getting the expected result. You'd probably want to subtract 1 and put the fractional difference desired into the tv_nsec field.

netniV:
Timedwait, or wait, basically sit there in the thread without any real way of knowing where it's paused or why.  This can make troubleshooting why spine locks up very tricky so I implemented and tested the trywait code (which does work just not long enough it seems).

The removal of that variable may be a good idea.  I'm guessing we would need to raise the 100, then use a mod of 100 for reporting purposes.  Even though that was tested, I'm guessing only the larger systems really notice the issue.

eriksejr:
I think the odds of any given user hitting this problem is inversely proportional to the number of threads they are running. If someone is running few threads and have a few a long running scripts the odds of spine needing to wait on the semaphore before creating another is going to be higher. In my case I work in a pretty high-latency environment for the network elements in question so even talking to some of these devices there's a 0.6s RTT for the packet alone, and that's just one packet. It's no wonder then that even with 64 threads a timeout of 1 seconds is too short for me.

I think the assumptions for the timeout being just under the poller interval in 1.2.17 were pretty reasonable. It allows for a use case where, for whatever reason someone has a situation where the poller is just executing a very small number of very long-running scripts.

If you like I can revert to the original code and play with the values a bit and let you know what works for me.

netniV:
I would prefer to have some form of log reporting on the threads.  I really don't mind the waiting game when it's required, but not knowing what thread is waiting and where, you can't really diagnose any issues.  Some people were often getting reports of a mismatch of thread counts under the old code and lockups.

The inverse rule makes sense because most of my devices respond pretty fast, not many are slow and it's likely that any that are, get handled because the rest are not.  Before I started my own company, one of my previous employers customers had devices quite remote so things like your scenario was very easy to hit and test against.

I can probably simulate it by creating a proxy that slows connections down or plug something into snmpd that runs a script that is slow at responding.

netniV:
So, I'm not convinced on this one.  There were some issues with the hostname parsing code that failed to allow for IPv6 names, and the reporting on the thread wait is too quick as it's actually half a second.  So I've corrected that now to be 5 seconds to show there's an outstanding thread.

I've run some more tests, setting host threads and other things but so far I haven't been able to break it in a way that needs changing.  Not saying that there isn't an issue somewhere, just I've been unable to replicate it.

netniV:
I managed to break things by changing the MAX_xx_SCRIPTS config.h setting to 1.  A bunch of changes have been made for instrumentation and extension on some timeouts to be a longer to give the system chance to catch up.  However, whilst these errors are now showing up, they would have been a problem anyway just you didn't know it as they would allow for indefinite timeouts on the semaphores.

If you can try the latest round of changes, I'd be interested to know if that fixes things.

eriksejr:
I will, i'll compile it all as it is right now and run it for a while. I'll let you know

eriksejr:
No issues so far with 2 completed poller runs. I built --with-max-scripts=25 since I do exceed the default of 20 by a bit:



```
2021/09/03 09:10:56 - SYSTEM DSSTATS STATS: Type:HOURLY, Time:0.3322 RRDUser:0.0000 RRDSystem:0.0000 RRDReal:0.0000
--
2021/09/03 09:10:55 - SYSTEM THOLD DAEMON STATS: TotalTime:0.012  MaxRuntime:0.012 Processed:2 InProcess:0 TotalDevices:573 DownDevices:0  NewDownDevices:0 MaxProcesses:1 Completed:1 Running:0 Broken:0
2021/09/03 09:10:55 - SYSTEM STATS: Time:53.3812 Method:spine  Processes:1 Threads:64 Hosts:573 HostsPerProcess:573 DataSources:12826  RRDsProcessed:0
```

netniV:
Yes, sorry, there's a new push so that the script waiting time is tied to the script server timeout value now (1.5x it for leeway).  The logging has been reduced too.

eriksejr:
Ok thank you, since it is working just fine for me now this PR does not seem necessary.

PR #216, opened 2021-07-08
Fix rrd_next_step not updated properly when rrd_step > poller_interval

eriksejr:
In my environment I have a 5 minute (300 second) poller interval which matches the rrd_step size of the vast majority of my graphs. However, I have one set of devices which only provides data every 15 minutes. As such I have created these graphs with a RRD template with a step size of 15 minutes (900 seconds) and a heartbeat of 20 minutes (1200 seconds). Cacti does not provide an option for a 15-minute polling interval (it jumps from 10 to 20 minutes) so I have patched global_arrays.php every time I have updated to add one.

This was all working fine until the update from 1.2.16 to 1.2.17 and then shortly after I updated to 1.2.18 but the problem remained. Looking into the problem deeper revealed that for the 900 second step devices, the poller was running only every 1200 seconds resulting in gaps in the graph periodically.

This issue will not appear on devices where the rrd_step == poller interval, so many setups will not see this problem. I don't believe it was a spine commit that broke this between the two versions - I'm guessing it was a cacti one? At any rate here is what is going on. In poller.c you have:

```
"UPDATE poller_item"
" SET rrd_next_step = IF(rrd_step = %i, 0, IF(rrd_next_step - %i < 0, rrd_step, rrd_next_step - %i))"
" WHERE host_id = %i", set.poller_interval, set.poller_interval, set.poller_interval, host_id);
```

The poller appears to only poll the device if rrd_next_step <= 0 or maybe rrd_next_step == 0 in the database though I did not verify the code on this, that is based on observation. The outer IF statement sets rrd_next_step to 0 any time the rrd_step is = poller interval, this prevents this issue from being seen on graphs where this is the case (probably the vast majority of graphs for most people). The problem is on the inner IF statement where rrd_step > poller_interval. Spine only updates rrd_next_step AFTER the poller run is complete. Decrementing this value BEFORE the run would be another way to solve this issue. Take my situation with arbitrary times beginning with a poller run:

rrd_step = 900
poller_interval = 300

Pass 1 - 13:00:
rrd_next_step = 0
Poller runs on device. Spine sets rrd_next_step = rrd_step (900).

Pass 2 - 13:05:
rrd_next_step = 900
Poller does not run on device. Spine sets rrd_next_step = rrd_next_step - poller_interval (900 - 300) = 600

Pass 3 - 13:10:
rrd_next_step = 600
Poller does not run on device. Spine sets rrd_next_step = rrd_next_step - poller_interval (600 - 300) = 300

Pass 4 - 13:15:
rrd_next_step = 300
Poller does not run on device. Spine sets rrd_next_step = rrd_next_step - poller_interval (300 - 300) = 0

Pass 5 - 13:20:
rrd_next_step = 0
Poller runs on device. Spine sets rrd_next_step = rrd_step (900)

First run 13:00 next run 13:20. Total duration between poller runs = 20 minutes. Correct duration between poller runs = 15 minutes.

The attached pull request corrects for this although you could simply remove the outer IF statement, as mathematically the result of the inner IF statement is no different in either scenario post-patch. This would make things slightly more efficient. I'm just concerned there could be some other reason it is there I am not aware of or seeing.


netniV:
Hmm, that may explain one or two issues seen on the forums regarding missing time periods in graphs. This will be another one for us to review.

netniV (changes requested):
Can you add a CHANGELOG for this and I can then get it committed.

eriksejr:
I screwed this up and merged my branch with develop to try to sort out the conflict with the CHANGELOG. I'll just re-open another PR for the changelog, sorry about that...

netniV:
Ah, you should just have pulled it.  In fact, if you revert to the above commit, you should be OK

netniV:
I think you should be able to do this by using the following:

```
git pull
git reset HEAD c908cc5f901fde3ab35859156ba1043b7708e4ea
git push -f
```

eriksejr:
Sorry I tried that but it didn't work so I created #228 . Wish I was better with git/github but learning....

netniV:
It's all a learning curve to get to somewhere your comfortable.  I thought the above might work as I was sure that's what I'd used in the past but it was all from my head so I may have made a mistake somewhere.  New PR is now committed, thanks!

PR #223, opened 2021-08-21
Fix non-void return warnings

andreasstieger:


netniV:
Thanks for the submission, but I'm going to close this as I believe we removed these functions since we never made use of them.

andreasstieger:
I do not understand - you say you believe you removed them.. but they are still there. Why else would the compiler be complaining? Can you find the reference for the removal?

If they are there but not used - should they be deleted outright?

TheWitness:
We need a bug filed.   If the functions are no longer in use, removing them would be prudent.  I've yet been able to do a thorough review of all @netniV changes. We had to in fact back off on one of them.

netniV:
My bad, you are comparing against develop, not the 1.2.x branch.  There may be some changes need porting over.

PR #225, opened 2021-09-01
Fixed: change 'restrict' to '__restrict__' for old gcc like 4.8.5

ddb4github:


netniV:
Thanks for the update, useful to know.  However, these functions were removed as we were not using them so I'm going to close this.

PR #242, opened 2021-11-05
Patch around potential buffer overflow within sql.c/db_escape()

### LINKED ISSUE #241: Invalid query errors when remote poller hosts return long sysDescription values
**Describe the bug**
sql.c/db_escape() does not properly consider the potential length of the string returned from `mysql_real_escape_string()`

**To Reproduce**
1. Configure Cacti using remote pollers, and spine as the polling mechanism
2. Add a host to a remote poller which returns a description containing a length near, but still under, 256 characters which also contains a lot of characters that require escaping.  Such as a Cisco version string containing multiple `\r\n` characters that is 250 characters in length like this:
3. Allow the remote poller to poll the host and try to synchronize host data back to the main instance

**Expected behavior**
I would expect this to complete without errors

**Server (please complete the following information):**
 - OS: Debian 10

**Compiling (please complete the following information):**

 - compiler: gcc (Debian 8.3.0-6) 8.3.0
 - autoconf: 2.69
 - glibc: 2,28-10
 - source: 1.2.19 release and latest 1.2.X branch (commit 616ae8d4fc5037643f046721a3cbb5f2871e335d)

**Additional Context**

I believe this is caused by the `sql.c/db_escape()` function not properly honouring the length recommendation provided in the MySQL C API documentation here:
https://dev.mysql.com/doc/c-api/8.0/en/mysql-real-escape-string.html

They specifically state that length is the number of characters processed from the input and that the buffer for the return must be equal to 2*n+1 bytes in length.

To fix this problem I propose that first, that length check is properly considered within `sql.c/db_escape()`, and second that the buffer within `util.c/poller_push_data_to_main()` is enlarged to hold 2*n+1 for the largest of input values, which is currently a varchar(300).

Assuming worst-case sizes for each value returned in this loop:
https://github.com/Cacti/spine/blob/616ae8d4fc5037643f046721a3cbb5f2871e335d/util.c#L823

The buffer `sqlbuf` would only hold approximately 550 rows.  This is uncomfortably close to the established limit so I propose we also raise the size of `sqlbuf` to help prevent innocent looking future changes from introducing hard to find bugs.

I am currently preparing a new branch and PR to submit as a resolution to this issue and will hopefully submit this shortly.

ctrowat:
This (hopefully) fixes #241

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ The Cacti Group | spine
 
 1.2.20
 -issue#238: Timeouts due to new Uptime checks
+-issue#241: Invalid query errors when remote poller hosts return long sysDescription values
 
 1.2.19
 -issue#172: Issues with polling loop may skip some datasources
--- a/sql.c
+++ b/sql.c
@@ -550,12 +550,10 @@ void db_escape(MYSQL *mysql, char *output, int max_size, const char *input) {
 	if (input == NULL) return;
 
 	char input_trimmed[BUFSIZE];
-	int  input_size;
+	int  max_escaped_input_size = (strlen(input) * 2) + 1;
 
-	input_size = strlen(input);
-
-	if (input_size > max_size) {
-		snprintf(input_trimmed, max_size - 10, "%s", input);
+	if (max_escaped_input_size > max_size) {
+		snprintf(input_trimmed, (max_size / 2) - 1, "%s", input);
 	} else {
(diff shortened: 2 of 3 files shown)


netniV (approved):

netniV:
I can't see an issue with this suggestion.  I'll merge and we can see if anyone runs into any further problems with it.

PR #245, opened 2021-12-10
configure.ac: fix bashisms

thesamesam:
configure needs to work with a POSIX-compliant shell so we need
to avoid bashisms like '=='. This fixes configure with e.g. /bin/sh
provided by dash.

Signed-off-by: Sam James <sam@gentoo.org>


netniV (approved):

TheWitness:
I prefer bash as my head hurts less when switching languages.  It's a minor point though.  Thanks for the contribution!

Issue #296, opened 2023-01-09
SIGSEGV, Segmentation fault after upgrade 1.2.7 to  1.2.23 


djpa3k:
Hi,

when i upgrade from 1.2.7 to 1.2.23 i get SIGSEGV when is not posible get device uptime from OID .1.3.6.1.2.1.1.3.0 

```
(gdb) run -V 5 -R -C /usr/local/spine/etc/spine.conf
Starting program: /root/cacti-spine-1.2.23/./spine -V 5 -R -C /usr/local/spine/etc/spine.conf
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
SPINE: Using spine config file [/usr/local/spine/etc/spine.conf]
[New Thread 0x7ffff48ba700 (LWP 28428)]
[Thread 0x7ffff48ba700 (LWP 28428) exited]
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The path_php_server variable is /var/www/script_server.php
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The path_cactilog variable is /var/www/log/cacti.log
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The version variable is 5.5.68-MariaDB
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The log_destination variable is 1 (FILE)
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The path_php variable is /usr/bin/php
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The availability_method variable is 1
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The ping_recovery_count variable is 2
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The ping_failure_count variable is 2
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The ping_method variable is 2
2023/01/09 10:18:30 - SPINE: Poller[1] PID[28424] PT[140737354020992] DEBUG: The ping_retries variable is 3
[...]

PR #297, opened 2023-01-09
Fix possible SIGSERV error

djpa3k:
#296  - return NULL from snmp function could cause an error in some environments


netniV (changes requested):
Can you add a CHANGELOG entry for this too? Thanks!

TheWitness:
Good catch.

TheWitness (approved):
Looks good

netniV (approved):
Although this targets 1.3.0, I can cherry pick this back to the 1.2.x branch as we need it in both.

PR #306, opened 2023-04-04
bootstrap shell script contains some PHP cruft

eriksejr:
Probably a copy/paste error, it does not impact the functionality of the script, just throws an error when it is run.


TheWitness:
@eriksejr, cool thanks for that.

Issue #324, opened 2023-11-07
Can't compile spine on OpenBSD


mikygee:
Hello,

I opened a bug in 2020 because I couldn't compile spine https://github.com/Cacti/spine/issues/163

Right now I cannot, I can compile it with spine 1.2.15 but cannot with spine 1.2.16
Of course spine 1.2.20 doesn't compile too

```
mv -f .deps/keywords.Tpo .deps/keywords.Po
egcc -DHAVE_CONFIG_H -I. -I./config      -I/usr/local/include/net-snmp -I/usr/local/include/net-snmp/.. -I/usr/local/include/mysql -g -O2 -MT error.o -MD -MP -MF .deps/error.Tpo -c -o error.o error.c
mv -f .deps/error.Tpo .deps/error.Po
/bin/sh ./libtool  --tag=CC    --mode=link egcc  -I/usr/local/include/net-snmp -I/usr/local/include/net-snmp/.. -I/usr/local/include/mysql -g -O2   -L/usr/local/lib -L/usr/local/lib  -o spine sql.o spine.o util.o  snmp.o locks.o poller.o  nft_popen.o php.o ping.o  keywords.o error.o  -lnetsnmp -lmysqlclient -lm -lmysqlclient -lm -lcrypto -lz -lpthread -lm -lpthread -lssl
libtool: link: egcc -I/usr/local/include/net-snmp -I/usr/local/include/net-snmp/.. -I/usr/local/include/mysql -g -O2 -o spine sql.o spine.o util.o snmp.o locks.o poller.o nft_popen.o php.o ping.o keywords.o error.o  -L/usr/local/lib -lnetsnmp -lmysqlclient -lcrypto -lz -lm -lpthread -lssl -Wl,-rpath,/usr/local/lib -Wl,-rpath,/usr/local/lib
sql.c:528(sql.o:(append_hostrange)): warning: sprintf() is often misused, please use snprintf()
util.c:1145(util.o:(die)): warning: vsprintf() is often misused, please use vsnprintf()
util.c:1151(util.o:(die)): warning: strcat() is almost always misused, please use strlcat()
util.c:1004(util.o:(poller_push_data_to_main)): warning: strcpy() is almost always misused, please use strlcpy()
ld: error: undefined symbol: backtrace
>>> referenced by error.c:68
>>>               error.o:(spine_signal_handler)

ld: error: undefined symbol: backtrace_symbols
>>> referenced by error.c:69
>>>               error.o:(spine_signal_handler)
[...]

PR #325, opened 2023-11-07
Correction bug #324 add -lexecinfo to configure.ac for OpenBSD

mikygee:
Added -lexecinfo to configure.ac so that spine can compile on OpenBSD


TheWitness:
Can you do this pull for 1.2.x instead?

mikygee:
I'm not confident with github manipulation.
Here's what I did
https://github.com/Cacti/spine/pull/326

PR #326, opened 2023-11-07
Correction bug Cacti#324 add -lexecinfo to configure.ac for OpenBSD

mikygee:
Correction bug https://github.com/Cacti/spine/issues/324 add -lexecinfo to configure.ac for OpenBSD


TheWitness:
I can't merge this.  It needs to be 1.2.x -> 1.2.x

PR #334, opened 2023-12-08
Fix Clang 16 compiler warning autological-pointer-compare

eriksejr:
While compiling sql.c clang 16 generates the following warning:

```
sql.c:407:27: warning: comparison of address of 'db_pool_local[id].mysql' not equal to a null pointer is always true [-Wtautological-pointer-compare]
                        if (&db_pool_local[id].mysql != NULL) {
                             ~~~~~~~~~~~~~~~~~~^~~~~    ~~~~
sql.c:428:28: warning: comparison of address of 'db_pool_remote[id].mysql' not equal to a null pointer is always true [-Wtautological-pointer-compare]
                        if (&db_pool_remote[id].mysql != NULL) {
                             ~~~~~~~~~~~~~~~~~~~^~~~~    ~~~~
```

If the conditional is always true then the fix is to remove the conditional in this instance

### DIFF
--- a/sql.c
+++ b/sql.c
@@ -404,15 +404,13 @@ void db_create_connection_pool(int type) {
 
 			db_connect(type, &db_pool_local[id].mysql);
 
-			if (&db_pool_local[id].mysql != NULL) {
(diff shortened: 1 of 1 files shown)


TheWitness:
I see an issue here I have to see about a connection failure.

TheWitness:
Okay, this is fine as the db_connect() will exit from spine if it fails.

PR #403, opened 2026-02-25
Fix POSIX sh compatibility in configure.ac

somethingwithproof:
## Summary

- Replace `[[ ]]` bash-isms with POSIX `[ ]` in MySQL header detection (lines 239, 245)
- Add proper variable quoting to prevent word splitting
- Use `-n` instead of `! -z` because life's too short for double negatives

`./configure` currently fails on Debian/Ubuntu where `/bin/sh` is `dash`, not `bash`. Two lines, zero risk.

## Test plan

- [ ] Run `./configure` on a Debian/Ubuntu system (dash as /bin/sh)
- [ ] Run `./configure` on a system with bash as /bin/sh (no regression)
- [ ] Verify MySQL headers are still detected correctly

Signed-off-by: Thomas Vincent <thomasvincent@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -236,13 +236,13 @@ AC_DEFUN([MYSQL_LIB_CHK],
 MYSQL_SUB_DIR="include include/mysql include/mariadb mysql";
 for i in $MYSQL_DIR /usr /usr/local /opt /opt/mysql /usr/pkg /usr/local/mysql; do
   for d in $MYSQL_SUB_DIR; do
-    if [[ -f $i/$d/mysql.h ]]; then
+    if [ -f "$i/$d/mysql.h" ]; then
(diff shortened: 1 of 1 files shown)


somethingwithproof:
Closing this PR — the change is incorrect for autoconf.

In `configure.ac`, `[[ ]]` is the correct M4 idiom, not a bashism. M4 treats `[` and `]` as quote delimiters, so `[[ -f $path ]]` in `configure.ac` produces `[ -f $path ]` in the generated `configure` script (valid POSIX). Replacing with `[ -f "$path" ]` causes M4 to strip both brackets, producing bare `-f "$path"` — which fails with `/bin/sh: -f: command not found`.

Verified by running `autoconf` on both versions. The existing code is correct.

PR #407, opened 2026-02-27
fix: strip -Werror=declaration-after-statement inherited from net-snmp-config

### LINKED ISSUE #406: net-snmp 5.9.5 injects -Werror=declaration-after-statement; ~20 C89 violations in Spine source
On FreeBSD (and potentially other platforms), net-snmp 5.9.5 ships a `net-snmp-config` that injects `-Werror=declaration-after-statement` into CFLAGS via pkg-config. Spine has ~20 places where variables are declared after statements (C99 style), which breaks the build under this flag.

This surfaced during the net-snmp 5.9.4 → 5.9.5 upgrade on FreeBSD 13.5/14.3. The SO version also bumped (.so.40 → .so.45), requiring a full Spine recompile.

### Affected files (partial audit)

**sql.c — `db_connect()`**
`char *ssl_key`, `ssl_ca`, `ssl_cert` declared inside `#ifdef HAS_MYSQL_OPT_SSL_KEY` after statements.

**sql.c — `db_escape()`**
`char input_trimmed[]` and `int max_escaped_input_size` declared after the `if (input == NULL) return;` guard.

**snmp.c — `snmp_host_init()`**
`size_t len` declared mid-block after the SNMPv3 auth/priv setup.

**util.c — `get_date_format()`**
`char log_sep = '/'` declared after the malloc guard.

At least ~16 more in other files (poller.c, snmp.c auth blocks, etc). Issue #109 fixed a similar class of violations in 2019 (for-loop initial declarations).

### Recommended fix

Set `-std=c11` in `configure.ac` (or the m4 macro that builds CFLAGS). C11 is supported by every compiler Spine targets:
- GCC 4.7+ (2012)
- Clang 3.1+ (2012)
- FreeBSD 10+ base compiler

This eliminates the declaration-after-statement issue permanently and avoids the maintenance burden of enforcing C89 style. It also unlocks `_Static_assert`, `_Alignof`, anonymous structs/unions, and `stdatomic.h` — all useful for Spine's threading and cache-line alignment.

The alternative (stripping the flag from inherited CFLAGS or enforcing C89 everywhere) is more fragile since new C99+ patterns will keep creeping in.

### FreeBSD port status

The FreeBSD port (maintained by skreuzer@FreeBSD.org) patched around this on Jan 18 2026. Setting `-std=c11` upstream would let them drop that patch.

Ref Cacti/cacti#6644, #109

somethingwithproof:
On FreeBSD, `net-snmp-config --cflags` injects `-Werror=declaration-after-statement` into the build environment. Spine uses C99 mixed declarations, so this flag breaks compilation.

Strip the flag early in `configure.ac` before any compilation checks run.

Tested with Docker (Debian bookworm, gcc 12):
- Baseline build: PASS
- Build with flag injected: PASS (flag stripped by configure)
- Without fix + flag injected: FAIL (confirms the bug)

Fixes #406

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,11 @@ AC_ARG_ENABLE(warnings,
   AC_MSG_RESULT(no)
 )
 
+# Strip -Werror=declaration-after-statement if inherited from net-snmp-config
+# on FreeBSD. Spine uses C99 mixed declarations; this flag breaks the build.
+# See https://github.com/Cacti/spine/issues/406
+CFLAGS=$(echo "$CFLAGS" | sed 's/-Werror=declaration-after-statement//g')
+
 AC_PATH_PROG(HELP2MAN, help2man, false // No help2man //)
 AC_CHECK_PROG([HELP2MAN], [help2man], [help2man])
 AM_CONDITIONAL([HAVE_HELP2MAN], [test x$HELP2MAN = xhelp2man])


somethingwithproof:
Superseded by #408 which fixes the root cause (moves declarations to block tops) instead of stripping the flag.

PR #418, opened 2026-03-01
fix: plug resource leaks in SNMP session setup and poll paths

### LINKED ISSUE #415: Resource leaks in SNMP session setup and poll paths
Four resource leaks found during code review:

1. **snmp.c SNMPv3 session setup**: `session.peername` assigned directly from the caller's pointer instead of `strdup()`. When `snmp_sess_open()` copies the session, the original pointer is not owned by the caller, leading to a dangling reference if the caller buffer is reused.

2. **snmp.c SNMPv3 OID leaks on error returns**: `securityAuthProto` and `securityPrivProto` allocated via `snmp_duplicate_objid()` are not freed on four early-return error paths (invalid auth protocol, invalid priv protocol, auth key generation failure, priv key generation failure). The happy path also leaked both OID allocations after `snmp_sess_open()` copied them.

3. **snmp.c PDU leak on OID parse failure**: `snmp_pdu_create()` allocates a PDU in `snmp_get_base()` and `snmp_getnext()`, but if `snmp_parse_oid()` fails, the function returns without calling `snmp_free_pdu()`.

4. **poller.c sprintf overflow in exec_poll()**: `sprintf(executable, "%s", proc_command)` with an unbounded source into a `BUFSIZE` buffer. Changed to `snprintf`.

somethingwithproof:
Fixes #415

Four resource leaks in snmp.c and poller.c:

1. snmp.c: `session.peername` not `strdup()`'d, dangling reference after `snmp_sess_open()`
2. snmp.c: `securityAuthProto`/`securityPrivProto` OIDs leaked on 4 error paths and happy path
3. snmp.c: PDU leaked in `snmp_get_base()` and `snmp_getnext()` when `snmp_parse_oid()` fails
4. poller.c: `sprintf` overflow in `exec_poll()`, switched to `snprintf`

All declarations at block top per project convention. Tested with gcc and clang.

### DIFF
--- a/poller.c
+++ b/poller.c
@@ -2288,13 +2288,13 @@ char *exec_poll(host_t *current_host, char *command, int id, char *type) {
 
 		/* peel the executable from the command */
 		saveptr = proc_command;
-		sprintf(executable, "%s", proc_command);
+		snprintf(executable, BUFSIZE, "%s", proc_command);
 		strtok_r(executable, " ", &saveptr);
 
 		/* cheesy little hack to add /usr/bin/ if its not included */
 		if (strstr(executable, "/") == NULL) {
 			saveptr = proc_command;
-			sprintf(executable, "/usr/bin/%s", proc_command);
+			snprintf(executable, BUFSIZE, "/usr/bin/%s", proc_command);
 			strtok_r(executable, " ", &saveptr);
 		}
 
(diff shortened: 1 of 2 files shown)


TheWitness (dismissed):

xmacan (approved):

bmfmancini (approved):

TheWitness:
@netniV, can you elevate Sean and Petr?  @thomasvincent , in the mean time do a pull and modify something so that you are the last commiter.

TheWitness (approved):

PR #452, opened 2026-03-12
fix: Incorrect buffer size

TheWitness:
The SSL Options point to files, and a reasonable path should not exceed 1024 bytes.  Additionally, we should comment out the cppcheck until we have had more time to review the warnings.  These are mostly, but not all style and non-fatal, but good to have addressed for optimal program flow.


somethingwithproof (commented):
.

somethingwithproof (commented):
.

somethingwithproof (commented):
.

somethingwithproof (commented):
.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
somethingwithproof: Latest runner failing?

somethingwithproof:
I could not push directly to `Cacti:ssl_buffers` from this environment, so I opened a tightly scoped companion PR against that branch: #453.\n\nIt contains only the in-scope `util.c` hardening changes for #452:\n- bounded config token parsing + malformed-line guard\n- safer capability-string assembly logic\n\nIf you prefer a direct commit on `ssl_buffers`, I can provide a clean cherry-pick command for commit `8edf029`.

netniV (approved):

PR #467, opened 2026-03-12
Hardening: Implement sql_buffer_t and refactor bulk queries

### LINKED ISSUE #466: SQL Layer Hardening: Implement sql_buffer_t and refactor bulk queries
## Summary

I’m planning a follow-up hardening pass for the Spine SQL layer. Currently, we build our bulk queries (like the main data push in util.c) using dozens of manual sqlp += sprintf(...) calls.

## Work
This approach is fragile for two reasons:
 1. It relies on manual pointer arithmetic and buffer tracking at every single step, which is a prime candidate for off-by-one errors or silent overflows if a field grows.
 2. It’s a maintenance headache—every time a field is added to the schema, we have to manually sync several blocks of repetitive sprintf logic.

The Plan:
 1. Implement a small sql_buffer_t helper that handles bounded snprintf and pointer advancement automatically (eliminating the manual += math).
 2. Refactor poller_push_data_to_main to use this helper. This will likely shrink util.c by ~100 lines and make the SQL structure much easier to audit and maintain.

Goal is to make the data path 100% 'safe-by-construction' and easier for us to add fields in the future without the boilerplate noise.

## Acceptance Criteria

- [ ] Work items are implemented and validated.
- [ ] Changes preserve behavior unless explicitly intended.

somethingwithproof:
Fixes #466.

## Summary
This PR implements SQL-layer hardening for bulk query construction by introducing a bounded SQL buffer helper and refactoring the main bulk insert path to use it.

## What changed
- Added `sql_buffer_t` in `sql.h`.
- Added helper functions in `sql.c`:
  - `sql_buffer_init()`
  - `sql_buffer_append()`
  - `sql_buffer_reset()`
- Refactored `poller_push_data_to_main()` in `util.c` to replace manual pointer arithmetic and repeated `sprintf` construction with bounded append operations.
- Keeps SQL query construction null-terminated and bounded by design.

## Scope
Rebuilt from `develop` to include only the SQL-layer hardening change for #466.


util.c
```diff
@@ -898,56 +900,53 @@ void poller_push_data_to_main() {
...
-						sqlp += sprintf(sqlp, " (");
+						sql_buffer_reset(&sb);
+						sql_buffer_append(&sb, "%s (", prefix);
 					} else {
-						sqlp += sprintf(sqlp, ", (");
+						sql_buffer_append(&sb, ", (");
 					}
 
-					sqlp += sprintf(sqlp, "%s, ", row[0]); // id mediumint
+					sql_buffer_append(&sb, "%s, ", row[0]); // id mediumint
```
netniV: Since we are introducing `sql_buffer_append` should we not also detect and report a failure to add to the buffer?

sql.c
```diff
@@ -625,4 +625,62 @@ int db_column_exists(MYSQL *mysql, int type, const char *table, const char *colu
...
+ * Returns the number of characters written, or -1 on overflow.
+ */
+int sql_buffer_append(sql_buffer_t *sb, const char *format, ...) {
+	va_list args;
+	int     remaining;
+	int     written;
+
+	remaining = (int)sb->size - (int)(sb->ptr - sb->buffer);
+	if (remaining <= 0) {
+		return -1;
```
netniV, line 690: Do we report here a failure to be abler to append the buffer? Should it also be fatal @TheWitness ?

sql.c
```diff
@@ -625,4 +625,62 @@ int db_column_exists(MYSQL *mysql, int type, const char *table, const char *colu
...
+	}
+
+	va_start(args, format);
+	written = vsnprintf(sb->ptr, remaining, format, args);
+	va_end(args);
+
+	if (written < 0 || (size_t)written >= (size_t)remaining) {
+		/* Overflow or error - Ensure string is null-terminated */
+		sb->ptr[0] = '\0';
+		return -1;
```
netniV, line 706: Do we report here a failure to be abler to append the buffer? Should it also be fatal @TheWitness ?

netniV (changes requested):

TheWitness:
In my RTM work years ago, I wrote a linked list approach where you would dynamically allocate more memory as the buffer grew.  There were three main functions as memory serves me, but I could be wrong.  It was designed for bulk upserts as there was just so much data updated every few seconds...

1. Create a sql prefix as a string - "INSERT INTO (blah, blah, blah) VALUES "
2. Create a sql suffix - " ON DUPLICATE KEY UPDATE ...."
3. Assign a pointer to the item linked list for each bulk insert class/table
4. Create a function to add items to the buffer, they were not prepared, but quoted as required in the format: "(a, b, c, d, ...)".  You would pass in the pointer structure and the string, and the storage would be allocated and the string added to the buffer.
5. Create function that took the connection, the sql prefix, sql suffix, the pointer to the buffer to dump the link list with the appropriate max_packet_size to the database in chunks.
6. Create a function to freeup the linked list.

It's pretty close to this design actually.  But the buffer should be handled using malloc as it's extended and not a static buffer.  So, the only time you run into a memory problem is when the host runs out of it.  This is something that 'should' not happen in production.  But would need to be handled.  Something like a failsafe, if you run out of memory dump the buffer.

TheWitness:
So, if @somethingwithproof makes that design change, this is likely good.

```
// adds the sql_record to the beginning or end of the list
addRecord(connection, last_insert_pointer, top_pointer, sql_record);

// bulk writes as the prescribed/defined max_packet_size.  
// Often times this max_packet_size is NOT the MariaDB/MySQL 
// max packet size as we found that upserts are fastest when the 
// transaction sizes are on the order of about 128k bytes
writeToDatabase(connection, sql_prefix, sql_suffix, top_pointer);

// free the entirety of the linked list
freeupRecords(top_pointer);
```

I believe the link list was uni-directional.

TheWitness:
The reason I did bring this approach to spine was that it's basically two tables with the same structure and two connections.  So, the value of adding the overhead to me was much less.  We would always dump and reset the buffer after each bulk write.

TheWitness:
We are doing basically the same thing in the PHP code.  With PHP, it's just so much easier as it's my preferred language.

somethingwithproof:
Initial implementation details for `dc783f1` (later tightened by follow-up `1b513a5`):

- `sql_buffer_append()` fast path now attempts write into remaining capacity first and reallocates only on truncation.
- Added max-cap enforcement via `SQL_MAX_BUFFER_CAPACITY` and explicit logging for invalid args/state/format/cap/realloc failures.
- `poller_push_data_to_main()` now handles append failures per-row with flush+single-retry behavior, then hard-logs and drops only that row if retry still fails.

For the latest consolidated status and rationale (including response to @TheWitness), see the updated follow-up comment below.

sql.c
somethingwithproof: Good call — this path now reports explicitly. In dc783f1, `sql_buffer_append()` logs on invalid args/state/format failure/cap breach/realloc failure before returning `-1`. In 1b513a5 I also tightened the state guard to `length >= capacity` to fail closed before any write attempt.

util.c
somethingwithproof: Addressed in dc783f1. `poller_push_data_to_main()` now checks append return codes via row-level helpers, logs failures, flushes/retries once, and if retry still fails logs a hard error and drops only that row. This avoids emitting malformed partial SQL while preserving the rest of the poller run.

sql.c
somethingwithproof: This truncation/growth path now reports and propagates errors as well (dc783f1): failed append operations are logged in `sql_buffer_append()` and returned to callers. Caller policy in `util.c` is row-level fail-safe (flush+single retry, then drop row) rather than process-fatal, so one bad row does not abort the full poll cycle.

somethingwithproof:
Update after follow-up commit `1b513a5` (building on `dc783f1`).

What changed:
- `sql_buffer_append()` now fails closed when buffer state is invalid (`length >= capacity`) before formatting.
- Append/growth failures are logged and propagated consistently.
- `poller_push_data_to_main()` keeps row-level fail-safe behavior: flush+retry once, then hard-log and drop only that row if retry still fails.
- Existing guardrails remain: max buffer cap and prevention of malformed partial SQL batches.

Response to @TheWitness:
- I kept this PR on the bounded dynamic-buffer approach to keep the patch surgical and reduce regression risk.
- The linked-list writer model you described is valid and could be a good follow-up refactor if we want to optimize batching behavior further.

SQL buffer failures are now handled loudly and safely, bad rows are isolated, and the poller avoids emitting broken SQL batches.

somethingwithproof:
I've pushed a follow-up commit to cap `syslog` output at 1024 characters in `spine_log()` as suggested by @TheWitness to prevent issues with large log buffers.

somethingwithproof:
Closing to resubmit with standardized branch name: issue-466-sql-layer-hardening

PR #493, opened 2026-03-12
Fixing spine capabilities

TheWitness:
In my SQL buffers commit, I had broken spine capabilities.  Should learn not to code after a long days work and visiting with loved ones.

### DIFF
--- a/util.c
+++ b/util.c
@@ -780,7 +780,6 @@ void read_config_options() {
 	// cppcheck-suppress knownConditionTrueFalse
 	strcat(spine_priv, (strlen(spine_priv) > 0 ? ",AES256":"AES256"));
 	#endif
-	strcat(spine_priv, "\"");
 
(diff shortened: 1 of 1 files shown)


somethingwithproof:
Nice cleanup here — splitting auth/priv into separate buffers makes this much easier to follow.
I do see one edge case though: when NETSNMP_DISABLE_MD5 is enabled, we still append ",SHA" because strlen(spine_auth) > 0 is true after the prefix, so output becomes authProtocols: ",SHA" (leading comma). Could we switch to a has_auth_algo flag for delimiter handling? Same idea on the priv side to avoid the > 16 magic-length check (has_priv_algo would be clearer/safer)?

TheWitness:
Let me do a small re-write and fix the one bug identified by Copilot.

TheWitness:
Okay, that's better.

netniV (approved):

PR #521, opened 2026-03-26
fix(ping): eliminate seteuid race by pre-opening ICMP socket at init

somethingwithproof:
## Problem

`ping_icmp()` called `seteuid(0)` inside each worker thread to open a raw ICMP socket. `seteuid` is process-wide, so while one thread held `LOCK_SETEUID` with euid=0, other threads that did not check the mutex could inherit the elevated privilege. The mutex serialized the seteuid calls but did not close the privilege window for concurrent threads.

## Fix

Open the ICMP raw socket once in `spine.c main()`, before any worker threads start. The single seteuid(0)/seteuid(getuid()) pair now happens in a single-threaded context with no race.

`ping_icmp()` calls `dup(icmp_socket)` to get a per-thread fd, so `select()`, `setsockopt()`, and `close()` in each thread operate on independent descriptors without interfering with other threads or the global socket.

If the socket cannot be opened at startup (permissions, capability check), `icmp_avail` is set to `FALSE` and spine falls back to UDP ping as before.

## Changes

- `spine.c`: add `int icmp_socket = -1` global; open raw socket before thread pool starts
- `spine.h`: `extern int icmp_socket`
- `ping.c`: remove all `seteuid`/`LOCK_SETEUID` blocks from `ping_icmp`; replace local socket open loop with `dup(icmp_socket)`; rename local `icmp_socket` to `icmp_fd` throughout

## Verification

Built clean on macOS aarch64 (gcc -std=gnu23, no errors or warnings).
Net change: 49 insertions, 103 deletions.


somethingwithproof:
Consolidated into mega PR #522 for independent mergeability.

PR #525, opened 2026-04-08
fix(ping): correct inverted strncasecmp in get_namebyhost

somethingwithproof:
strncasecmp() returns 0 on match, but the comparisons used the raw return value as truthy, inverting the logic. TCP matched non-TCP strings and vice versa. Also fixes comparisons against 'hostname' instead of 'token', and strncasecmp length 3 for 4-char strings TCP6/UDP6.


somethingwithproof:

somethingwithproof:

PR #526, opened 2026-04-08
fix(cli): save --mode argument before comparing to avoid argv advance

somethingwithproof:
getarg(opt, &argv) advances the argv pointer on each call. Three successive calls in the --mode handler consumed three argv entries instead of one, corrupting subsequent argument parsing when using the space-separated form (--mode online).


somethingwithproof:

somethingwithproof:

PR #527, opened 2026-04-08
fix(cli): check strchr return for NULL before dereference in -O handler

somethingwithproof:
When -O is passed without a colon separator, strchr returns NULL. The next line dereferences it unconditionally, causing a segfault. Add NULL check before dereference.


somethingwithproof:

somethingwithproof:

PR #528, opened 2026-04-08
fix(log): add missing break statements in get_date_format switch

somethingwithproof:
Every case in the log_datetime_format switch fell through to the next, so all date format options produced the default format regardless of configuration.


somethingwithproof:

somethingwithproof:

PR #530, opened 2026-04-08
fix(db): correct inverted return value in putsetting

somethingwithproof:
db_insert returns TRUE on success and FALSE on failure. putsetting checked 'result == 0' and returned TRUE, inverting the semantics. Callers received success on insert failure.


somethingwithproof:

somethingwithproof:

PR #534, opened 2026-04-13
fix(poller): add NULL check after db_get_connection at poll entry

somethingwithproof:
Follow-up to #531 which fixed the same NULL-dereference pattern at a different call site in poller.c.

## Problem

`db_get_connection(LOCAL)` at poller.c:235 can return NULL after a connection pool error (sql.c:525). The code immediately dereferences `local_cnn->mysql` without checking, causing a segfault when the pool is exhausted or the database is unreachable.

The same pattern exists for the `db_get_connection(REMOTE)` call at line 239.

## Fix

Add NULL checks for both LOCAL and REMOTE connections before dereferencing. On LOCAL failure, return immediately. On REMOTE failure, release the LOCAL connection before returning to avoid a pool leak.

Identified during code review of the spine PR batch (#525-#532).

### DIFF
--- a/poller.c
+++ b/poller.c
@@ -231,12 +231,21 @@ void poll_host(int device_counter, int host_id, int host_thread, int host_thread
 	MYSQL_RES *result;
 	MYSQL_ROW row;
 
-	//db_connect(LOCAL, &mysql);
+	/* db_get_connection can return NULL after a connection pool error */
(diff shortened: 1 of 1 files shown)


somethingwithproof:
Superseded by #523. The null-connection poller fix from this PR is already included there, so review/merge should continue on #523 only to avoid split review and conflict risk.

PR #537, opened 2026-04-15
[DRAFT 2.0] spine: TLS-by-default and plugin-contract behavior changes

somethingwithproof:
Carved out from #535 per @TheWitness — behavioral changes that affect the plugin state-machine contract (thold, amqp-publisher, boost) land in 2.0, not 1.3.

### A1. TLS-by-default for MySQL (6896444)
Flips `DB_UseSSL` / `RDB_UseSSL` default from `0` to `1` (preferred, with graceful fallback). Plugin-observable via connection setup timing and cleartext-required MySQL servers rejecting the handshake.

**1.3 behavior (matrix branch):** default remains `0`, operators opt in via spine.conf.
**2.0 behavior (this branch):** default is `1`, operators opt out explicitly.

## Why these items defer

Each change touches one of the plugin contract layers:
- `poller_output` write timing or content
- `poller_time` row shape
- `host.status` transition semantics
- Default connection / IPC behavior visible to plugins

## State-machine RFC

Planning an RFC doc to define the 2.0 contract with @TheWitness:
1. State machine as it exists today (`poller_output`, `poller_time`, `host.*`)
2. What each plugin relies on (thold, amqp-publisher, boost, weathermap, syslog, snmpagent, mactrack)
3. Migration surface (deprecation warnings in 1.3 upgrade script)
4. 2.0 daemon architecture (event-loop, in-process plugin hooks, streaming output)

## Next steps

- [ ] Open RFC PR for state-machine contract doc
- [ ] Implement 1.3 metachar-warning script (`cacti_check_shell_metachars.php`) for deprecation window
- [ ] Plugin compatibility matrix review

Closes nothing on 1.3; tracking for 2.0 only.

Related: #523 (CMake migration, 1.3), #535 (platform + security + operator tooling, 1.3).


somethingwithproof:
Addressed the outstanding security-analysis findings in dd8550f:

- moved workflow inputs and GitHub context values into quoted environment variables;
- split apt packages into an argv array before invoking apt-get;
- made test null checks control flow before dereferences;
- replaced string-literal boolean assertions with an explicit message assertion;
- removed the unused PHP parser assignment and unreachable ICMP condition; and
- handle ICMP send failures instead of overwriting the result.

Validation: actionlint and workflow policy pass; the reported cppcheck patterns are gone; CMake build passes; all 19 CTest tests pass.

somethingwithproof:
Closing in favour of #523.

These branches are not three features. All three independently delete autotools (`configure.ac`, `Makefile.am`, `Makefile.in`, `bootstrap`), move the sources into `src/`, and add CMake — and they conflict with each other across 18 to 69 files. This branch shares 89 of #523's commits, so it is a fork of an older #523 head rather than independent work.

The TLS work here is worth keeping and I will bring it back separately: the `bool`/`my_bool` width fix at the `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` boundary, the missing `mysql_ssl_set()`, and the `mysql_get_ssl_cipher()` downgrade check are bug fixes that belong on 1.2.x, distinct from the 2.0 default flip.

PR #538, opened 2026-04-17
security: critical race, spine.conf policy, signal handler, fd leaks

somethingwithproof:
Six fixes from the recent security review (see internal report, 2026-04-17). Each is one commit with a matching unit test.

## Scope

- **C1** (`poller.c`): guard `kill(pid, SIGKILL)` against `nft_pchild` returning -1. Without the guard, `kill(-1, SIGKILL)` wipes every process owned by the spine uid when the pid lookup fails (reachable from the select-timeout branch racing cleanup).
- **H1** (`util.c`): tighten spine.conf validation to match `SECURITY.md`. Refuse start on world-readable, group-readable, or owner-mismatch. Open config with `O_NOFOLLOW|O_CLOEXEC`. Previous behavior warned and continued.
- **H2** (`error.c`): rewrite fatal signal handler to async-signal-safe primitives. Pre-formatted messages + `write(STDERR_FILENO, ...)` + `_exit(128+signo)`. Drops SIGPIPE from the fatal list and ignores it process-wide.
- **H3** (sockets, MySQL, audit, log): set `FD_CLOEXEC` so poll-script children do not inherit the raw ICMP socket, MySQL session, audit netlink, or log fd.
- **H8** (`ping.c`): fix 1-byte OOB write in `get_namebyhost` when the hostname token equals `sizeof(name->hostname)`. Uses existing `strncopy` helper.
- **H9** (`snmp.c`): carry SNMPv3 contextEngineID as binary with explicit length. Previous `strlen(engine_id)` path truncated at the first `0x00`, weakening USM key localization.

## Verification

`cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug && cmake --build build && ctest --test-dir build --output-on-failure` — 24/24 pass. New tests: `test_poller_pid_guard`, `test_config_perms`, `test_signal_handler`, `test_cloexec`, `test_get_namebyhost`, `test_snmp_engine_id`.

## Risk

- H2 removes SIGPIPE as a fatal signal. Callers that rely on EPIPE to propagate via handler must check write return codes. Grep confirms no such caller exists in-tree.
- H1 makes previously tolerated 0640/0644 spine.conf fatal. Operators on shared-group deployments must `chmod 0600`.


somethingwithproof:
Superseded by consolidated PR #535 per maintainer's precedent in #523 (which explicitly cites merge-order conflict risk across util.c / spine.c / platform / CI files). All commits (C1 kill guard, H1 config perms, H2 signal handler, H3 FD_CLOEXEC, H8 strncopy, H9 engine ID binary length) are now on feat/distro-test-matrix via fast-forward. No content lost.

PR #539, opened 2026-04-17
fix(config): length-bounded tokenizer, range checks, umask, credential scrub

somethingwithproof:
Three fixes for config parsing and startup hygiene.

## Scope

- Replace `sscanf("%15s %255s", ...)` in `read_spine_config` with a length-bounded tokenizer. Previous parser silently truncated keywords > 15 bytes, values > 255 bytes, and passwords with spaces. Adds range checks for `DB_Port`, `RDB_Port` (1-65535), `CircuitBreakerThreshold` (1-1000000), and `DB_UseSSL`/`RDB_UseSSL` (0 or 1). Opens spine.conf with `O_NOFOLLOW|O_CLOEXEC`.
- `umask(027)` at startup before any file creation.
- Extract `spine_scrub_secrets()` and call from both the clean-exit path and `die()`. Uses `explicit_bzero` where available.

## Verification

21/21 tests pass. New tests: `test_config_parser` (keyword length, spaces in value, overlong line, embedded NUL, port range, symlink), `test_startup_umask`, `test_scrub_secrets`.

## Risk

- Operators with non-standard config files that rely on the prior silent truncation will now see a clear rejection. This is the intended behavior.


somethingwithproof:
Superseded by consolidated PR #535. Config parser tokenizer + range checks + umask + credential scrub are now on feat/distro-test-matrix via fast-forward.

PR #540, opened 2026-04-17
fix: memory and concurrency cleanup

somethingwithproof:
Memory and concurrency cleanup from the security review.

## Scope

- `log`: initialize `flogmessage` before `strftime` to prevent reading uninitialized stack on strftime failure.
- `util`: guard `die()` `strncat` against size underflow; replace `strcat` in auth/priv list builder with bounded `snprintf`-append; guard `snprintf` accumulators against truncation in `append_hostrange` and related sites.
- `breaker`: stack-copy state before unlock so logging survives a future concurrent delete; clamp `CircuitBreakerThreshold` to `[1, 1000000]`.
- `nft_popen`: replace `assert(prev != NULL)` with a runtime null check for concurrent unlink (safer under `NDEBUG`).
- `db`: null-guard `mysql_fetch_row` dereferences at six hot-loop call sites.
- `php`: break `readpipe` loop when the result buffer is full (prevents 0-size read spin).
- `log`: redact SNMP community strings and auth args (`-c`, `-u`, `-a`, `-x`, `-p`, `--password`, `--secret`) from command-error logs.
- `signals`: set `sa.sa_flags` explicitly per sigaction block.
- `poller`: null-guard `details[device_counter]` in the worker thread.

## Verification

21/21 tests pass. New tests: `test_snprintf_guard`, `test_redact_args`; extended `test_circuit_breaker` with clamp case.


somethingwithproof:
Superseded by consolidated PR #535. All 11 memory/concurrency cleanup commits are now on feat/distro-test-matrix via fast-forward.

PR #543, opened 2026-06-10
harden: validate ICMP reply identity and use _exit in fatal handler

somethingwithproof:
Two defensive changes split out from the bug-fix PR because they change
runtime behavior rather than fixing an outright defect.

Stacked on #542 (the ICMP length check lands there first). This PR's diff
will show #542's commits until that merges; the hardening-only commits are
the two listed below. Merge #542 first, then this rebases to two commits.

- ping.c: accept an ICMP echo reply only when its id and sequence match
  the request this thread sent. Spine uses a shared raw socket, so a reply
  from any host with the right source address could otherwise mark a
  device up. The request already sets id = pid & 0xFFFF and an
  incrementing sequence; this checks them on the way in.
- error.c: the SIGSEGV handler called exit(1), which runs atexit and stdio
  flush handlers from signal context. Switched to _exit(1).

## Risk
The ICMP id/seq check changes which replies count as "up." Needs a
reachability smoke test on real devices before merge.

## Verification
Compiles clean (cc -fsyntax-only -Wall, 0 warnings). Not yet runtime-tested.

Depends on #542.

### DIFF
--- a/error.c
+++ b/error.c
@@ -73,7 +73,7 @@ static void spine_signal_handler(int spine_signal) {
 			break;
 		case SIGSEGV:
 			fprintf(stderr, "%s FATAL: Spine Encountered a Segmentation Fault\n", logtime);
-			exit(1);
+			_exit(1);
 			break;
 		case SIGBUS:
 			fprintf(stderr, "%s FATAL: Spine Encountered a Bus Error\n", logtime);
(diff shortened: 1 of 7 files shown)


somethingwithproof:
Folded into #542, which now carries these three commits plus fixes for the Cygwin `MSG_PEEK` spin (the peeked datagram was never drained, so one non-matching reply busy-spun until host timeout) and a debug log on the id/seq mismatch path.

Closing to keep the review to one branch.

PR #545, opened 2026-06-11
ci: add Cygwin 1.2.x build coverage

somethingwithproof:
## Summary

Adds a separate Cygwin/Windows support PR for the 1.2.x branch.

- add a Windows-hosted GitHub Actions workflow that installs 64-bit Cygwin dependencies, runs `bootstrap`, configures against `/usr`, and builds with verbose output
- make `configure.ac` explicitly support only x86_64 Cygwin and fail clearly for unsupported 32-bit Cygwin hosts
- clarify documentation that Spine 1.2.x supports Cygwin on Windows, not native Windows builds
- preserve existing `CYGWIN` environment options while adding `nodosfilewarning` when missing
- store the selected Cygwin shell path explicitly instead of using an integer location flag
- use `select(fd + 1, ...)` in poller/ping paths instead of `FD_SETSIZE`

## Validation

- `git diff --check`
- workflow YAML parse with Ruby `YAML.load_file`
- local C syntax check with installed MySQL/MariaDB, Net-SNMP, OpenSSL headers and `-DHAVE_LIBPTHREAD=1`

The actual Cygwin build is expected to run in GitHub Actions after this PR is opened.

### DIFF
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -21,6 +21,13 @@ jobs:
     timeout-minutes: 30
 
     steps:
+      # windows-latest defaults to autocrlf=true, which gives the shell scripts
+      # CRLF endings that Cygwin bash cannot parse. Force LF before checkout.
+      - name: Use LF line endings
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
+
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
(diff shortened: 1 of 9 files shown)


TheWitness (dismissed):

somethingwithproof:
Folded into #544. The Cygwin job exists to validate the `fork()`/`FD_CLOEXEC` changes in that PR, so they belong together.

The out-of-scope portability work that was riding along here (the `configure.ac` 32-bit rejection, the `cygwin_sh_path` struct change, and the `select(fd + 1)` edits) is split out to `split/cygwin-portability-1.2.x` and will come as its own PR.

PR #548, opened 2026-07-14
chore: standardize enterprise Dependabot configuration

somethingwithproof:
Adds grouped, cooled-down Dependabot updates with standard ecosystem labels and commit prefixes.


somethingwithproof:
Folded into #550, which now carries the Dependabot config alongside the Rocky Linux job. Both are CI plumbing and read as one decision.

Closing this one.

PR #555, opened 2026-08-30
fix(ping): ping IPv6 devices with ICMPv6 echo requests

### LINKED ISSUE #127: IPv6 ping problem
I have host with IPv6 address (not DNS name) like 2a02:xxxx:xxx:1::face. With Cacti issue #3141 I can ping ipv6 address in gui. Spine has problem with ipv6 address:

spine -R -H 669 -V 10
2019/12/12 09:56:26 - SPINE: Poller[1] Device[669] DEBUG: Entering ICMP Ping
2019/12/12 09:56:26 - SPINE: Poller[1] Device[669] PING Result: ICMP: Destination hostname invalid

Spine 1.2.5, FreeBSD 11.x

 - compiler: clang
 - autoconf: autoconf 2.69_3
 - glibc: 2.56
 - source: release

### LINKED ISSUE #250: Support kernel parameter ping_group_range allowing unprivileged users to use ICMP (ping)
Today SPINE is only supporting ICMP if running as root -
to verify that it has root permissions capabilities for RAW_NET (hasCaps()) as well as setuid will be taken into account.

Since kernel release 2.6.x parameter ping_group_range will be supported that allows a range of user groups to use RAW sockets for PING as unprivileged users.

Many distributions are supporting that today per default, OpenNMS is also mentioning that option.
( see [https://opennms.discourse.group/t/how-to-allow-unprivileged-users-to-use-icmp-ping/1573] )

The parameter vvalue can be reviewed at CLI level for example with:

`browniebraun@browniebraun-Lenovo-ideapad-530S-14IKB:~$ /sbin/sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 0	2147483647
` 
If disabled net.ipv4.ping_group_range will return "1    0".

ping.c offers a function called inet_get_ping_group_range_net(), probably this could be used as an entry point for a modification of Spine.

somethingwithproof:
Spine opens its ICMP socket as `AF_INET`/`SOCK_RAW` only, so a device with an IPv6 address never reaches a working code path and reports `ICMP: Destination hostname invalid`. Reported in #127 in December 2019.

This adds an ICMPv6 echo path and dispatches to it in `ping_host()` when the address resolves to IPv6 and the ping method is ICMP. IPv4 is untouched: `ping_icmp()`, `ping_udp()` and `ping_tcp()` are unmodified, and IPv6 with a UDP or TCP ping method still falls through to the existing message.

Verified in a clean ubuntu:24.04 container at 39 warnings, matching develop.

**Evidence.** The ping path was extracted verbatim into a harness that stubs only the spine runtime, then run against real addresses:

```
target=::1                     result=HOST_UP   status=0.02408 response=ICMPv6: Device is Alive
target=2a02:0db8:0abc:1::face  result=HOST_DOWN status=down    response=ICMPv6: Ping timed out
target=notahost.invalid        result=HOST_DOWN status=down    response=ICMPv6: Destination hostname invalid
```

The second line is the reporter's exact case: the literal now produces a real echo request and an honest timeout instead of being rejected as an invalid hostname.

**Portability.** Everything sits behind `SPINE_HAVE_ICMPV6`, defined only when `AF_INET6`, `IPPROTO_ICMPV6`, `ICMP6_FILTER`, `ICMP6_ECHO_REQUEST` and `ICMP6_ECHO_REPLY` are all present and the target is not Cygwin, matching how `common.h` already excludes `<netinet/icmp6.h>` there. Without it the file compiles exactly as before. Four `configure.ac` checks were added for the scope-id helper (`getifaddrs`, `if_nametoindex`).

**Note for #523.** The version of this code on that branch sets `sin6_port` to 7 on the destination `sockaddr_in6`. On a raw IPv6 socket that field is the protocol number, not a port, so Linux rejects every `sendto()` with `EINVAL` — it fails on the first send. Fixed here by setting it to 0. The same bug is on the `feat/spine-2.0-behavioral` branch.

Closes #127

**Added in this revision**
- ICMP was gated on root or `cap_net_raw` in two places: `set.icmp_avail` in `util.c`, which diverts to UDP before `ping_icmp()` runs, and the socket call itself. Both now try `SOCK_DGRAM` first and fall back to the raw socket, so `net.ipv4.ping_group_range` decides the outcome. The IPv4 receive path allows for the header the kernel strips on a datagram socket, and the IPv6 reply match drops the echo id, which the kernel reassigns, keeping sequence, source address and payload as the discriminators. `tests/integration/test_unprivileged_icmp.sh` covers all three privilege and sysctl combinations. Closes #250.

Depends on #542: `get_namebyhost` truncates its input by one byte on this branch, so `192.168.166.2` resolves as `192.168.166.` and no ICMP ping succeeds until that fix lands. Merge #542 first.


netniV (dismissed):

TheWitness (changes requested):
Merge conflicts.

somethingwithproof:
Rebased onto develop, conflicts resolved. `MERGEABLE`, `behind=0`.

The conflict came from #542 landing `spine_icmp_classify_reply()`, which this PR then has to work with. That function parses the IPv4 header and matches the echo id, and neither holds on an unprivileged datagram socket: the kernel strips the header and assigns the id itself. Added `spine_icmp_classify_dgram_reply()` alongside it, which checks length, type and sequence, and left the raw path calling the original unchanged.

Also restored `tests/fuzz/fuzz_icmp_reply.c`. It fell through the gap between #542 and #563 — it could not go with the test split because it calls `spine_icmp_classify_reply()`, and the rebuild of #542 carried source files only, so it ended up in neither. It is back here with an input byte selecting which classifier to drive, so both paths are fuzzed.

Verified in a clean `ubuntu:24.04` container: builds at 4 warnings matching develop, 39 cmocka assertions pass, both fuzz targets clean at 4000 runs each, and `tests/integration/test_unprivileged_icmp.sh` passes 10/10 across the three privilege and sysctl combinations.

somethingwithproof:
Conflicts are resolved; this merges cleanly against develop now.

somethingwithproof:
The merge conflict this was blocked on is gone. `feat/icmpv6-ping` is level with `develop`, zero commits behind, and a three-way merge against it reports no conflict:

```
$ git rev-list --count fork/feat/icmpv6-ping..upstream/develop
0
$ git merge-tree --write-tree --name-only upstream/develop fork/feat/icmpv6-ping
(clean)
```

GitHub agrees: `mergeable: MERGEABLE`. Nothing has been force-pushed here, so the branch is the same one that was reviewed; `develop` moved underneath it and the overlap resolved itself.

## One thing worth knowing before this is merged

It conflicts with #597, in `ping.c`. Not with `develop`, only with that branch:

```
$ git merge-tree --write-tree --name-only fork/feat/icmpv6-ping fork/consolidate/spine-batch
CONFLICT (content): Merge conflict in ping.c
```

Both change `ping_icmp()`. This one adds the ICMPv6 path; #597 gives the function a single cleanup label and removes an unnecessary `seteuid(0)` around `close()`, as part of fixing the packet leak in #593. `configure.ac` and `util.c` overlap too but auto-merge.

Whichever lands first, the other needs a rebase. I would rather that be this one, since it is the smaller rebase and #597 is the one carrying the larger diff. Happy to do it either way, or to rebase this onto #597 now if you would prefer them ordered.

somethingwithproof:
Coordination update: dependency #542 merged on 2026-08-31, and GitHub now
reports this branch as mergeable against `develop`. The outstanding
changes-requested review only says "Merge conflicts"; that condition is no
longer present. Keep this PR separate and merge it before #597, which also
changes `ping.c`, `util.c`, and `configure.ac`.

somethingwithproof:
Rebased onto develop. Conflicts are resolved and the branch shows mergeable.

xmacan (approved):

somethingwithproof:
Consolidated into #619 to avoid reviewing and rebasing the same ping.c/util.c state machine twice. #619 now includes ICMPv6, unprivileged ICMP sockets, resolver/resource ownership, privilege-safe retries, fuzzing, and integration coverage. No implementation was dropped.

somethingwithproof:
Superseded by #619.

PR #557, opened 2026-08-30
port: memory-safety fixes from Spine #544 to develop

somethingwithproof:
Forward-port of the applicable fixes from #544, rebuilt on develop rather than carried across, so the obsolete release and CI history in #544 is left behind. 12 files, +269/-38.

**Out-of-bounds reads and writes**

- `ping.c` (`ping_icmp`): a raw-socket reply was cast to `struct ip` and the ICMP header located at `ip_hl << 2` with no length check, so a short or malformed reply read past `socket_reply`. The reply is now dropped unless it holds a full IP header, the computed header length is at least `sizeof(struct ip)`, and the buffer reaches the end of the ICMP header.
- `php.c` (`php_readpipe`): the read size was `RESULTS_BUFFER - (bptr - result_string)`, so a script server result that exactly fills the buffer puts the terminator one byte past it. The capacity now reserves the terminator, and an over-long result logs and returns undefined instead of truncating into the write. Offset arithmetic moved to `ssize_t`/`size_t` so it stays width-safe on LP64. (issue#561)
- `util.c` (`set_option`): `nopts` indexed `opttable` with no bound, so enough `-O` arguments walked off the end of the table. It now dies with an error.

**NULL dereferences**

- `spine.c`: `-O` parsing dereferenced the result of `strchr(setting, ':')` before testing it, so `spine -O foo` crashed instead of printing the usage error.
- `spine.c`: four `mysql_fetch_row()` results reached `atoi(row[0])` unchecked. The device loop breaks on NULL and the three count queries fall back to 0.
- `snmp.c` (`snmp_count`): `response->errstat` was read whenever the status came back `STAT_SUCCESS`, but Net-SNMP can report success with a NULL response. That case is now logged and the device treated as an error.
- `poller.c` (`poll_host`): the `poller_items` and `snmp_oids` allocations were used without checking `calloc()`. Both `die()` on failure now, matching the rest of the file.

**Shutdown could hang forever**

`php_close()` and `nft_pclose()` both reaped with a blocking `waitpid(pid, &status, 0)`. A wedged script server or command child held the caller in that call with no way out. Both now poll with `WNOHANG` for five seconds after `SIGTERM`, escalate to `SIGKILL`, wait one more second, then give up with a logged warning (`ETIMEDOUT` out of `nft_pclose`). `php_close()` also clears `php_pid` afterwards. This is the TODO that sat in `php.c` describing exactly this problem.

`error.c`: the `SIGSEGV` handler called `exit(1)`, running atexit handlers and stdio flushing inside an already-corrupt process. It calls `_exit(1)` now.

**Descriptors leaking into children**

The script server pipes (both directions) and the `nft_popen` pipe were created without `FD_CLOEXEC`, so every subsequent `exec` inherited them. All four are marked close-on-exec at creation, and the partially-created pipes are closed on the failure paths. `fcntl.h` added to `common.h`.

**The log date format never took effect**

`get_date_format()` was missing `break` in all six `switch` cases, so every configured format fell through to the default and Spine always logged `%Y-%m-%d`. `read_config_options()` was also reading only `default_datechar` into `log_datetime_separator`; it now reads `default_date_format` into `log_datetime_format` with `GD_*` bounds as well. Both were needed before the setting could do anything.

**Tests and CI**

`tests/regression/test_child_process_safety.sh` asserts the close-on-exec and non-blocking-reap invariants, so the blocking `waitpid` cannot come back unnoticed. CI now also runs on `1.2.x`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,34 +1,18 @@
 The Cacti Group | spine
 
-1.3.0
--issue#360: If a Remote Data Collector has not devices, spine does not properly register itself
--issue#423: Reinitialize fd_set before each select() call in ping_icmp() to prevent instant retry expiry after first timeout
--issue#424: Correct inverted spike_kill condition for PHP script server (IS_UNDEFINED -> !IS_UNDEFINED)
--issue#425: Mark all OIDs undefined when snmp_get_multi() receives errindex==0 from agent
--issue#426: Add NULL check after malloc in host_errors insert path to prevent UB on memory pressure
--issue#438: Resolve stale content in spine.c (--help search path), configure.ac (AC_PREREQ, obsolete macros), spine.conf.dist (missing settings), and Copilot instructions
--issue#441: Update copyright year to 2026 in 12 source files
--issue#445: Replace vfork/execv with posix_spawn in php_init; fix stack-local return in regex_replace
--issue#447: Correct off-by-one OOB write in strncopy() when src fills buffer; remove suppressing pragma
--feature#362: Report the number of data collection errors during data collection in host table
--feature#439: Add idiomatic RPM spec for cacti-spine (%autosetup, %config(noreplace), man1)
--feature#442: Add GitHub Actions CI: build matrix (gcc/clang), cppcheck, flawfinder, CodeQL
--feature#443: Add make targets for Docker build and verification (docker, docker-dev, verify, cppcheck)
--feature#444: Add multi-stage Dockerfile and Dockerfile.dev with ASan/cppcheck/scan-build
--feature#3740: Ability to disable a site
--feature#5090: Enhance number recognition within Spine
--feature#6001: Extend SYSTEM STATS
--feature: Make spine work with Stream feature
-
 1.2.31
--issue#365: Removing Backtrace Support
--issue#366: Failing to actually a poll device that returns "NoSuchObject"
--issue#370: Errors experienced with very long OIDs
--issue#373: Spine SNMPv3 SHA-512 auth doesn't work
--issue#374: Error DES is no longer supported on this system
--issue#383: Spine returns Segmentation fault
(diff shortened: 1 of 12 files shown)


CHANGELOG
```diff
@@ -1,34 +1,18 @@
...
--issue: Clear up a few very minor leaks in spine
+-issue: Bound PHP script-server shutdown and use width-safe pipe capacity arithmetic
+-issue#365: Removed Backtrace Support due to lack of OS support
+-issue#366: Prevent polling from stopping when sysDescr OID returns noSuchObject
+-issue#370: Increase buffer sizes to handle large OID numbers
+-issue#373: Spine cannot generate SNMPv3 Ku for SHA-384/SHA-512
+-issue#374: Improve depreciated DES support
+-issue#383: Improve Spine buffer handling #383
+-issue#384: Buffer overflow in ping_icmp() causes Spine crash
+-issue: Clear up some minor memory leaks
```
netniV: No tracker number for this

CHANGELOG
```diff
@@ -1,34 +1,18 @@
...
--feature#362: Report the number of data collection errors during data collection in host table
--feature#439: Add idiomatic RPM spec for cacti-spine (%autosetup, %config(noreplace), man1)
--feature#442: Add GitHub Actions CI: build matrix (gcc/clang), cppcheck, flawfinder, CodeQL
--feature#443: Add make targets for Docker build and verification (docker, docker-dev, verify, cppcheck)
--feature#444: Add multi-stage Dockerfile and Dockerfile.dev with ASan/cppcheck/scan-build
--feature#3740: Ability to disable a site
--feature#5090: Enhance number recognition within Spine
--feature#6001: Extend SYSTEM STATS
--feature: Make spine work with Stream feature
-
```
netniV, line 22: Changelog wiped out

netniV (changes requested):

netniV (approved):

Issue #567, opened 2026-08-31
get_date_format() rebuilds an invariant string, with a malloc/free pair, on every log line


somethingwithproof:
`spine_log()` calls `get_date_format()` once per log line, at `util.c:1477`. That function allocates, branches and formats every time:

```c
char *get_date_format(void) {
	char *log_fmt;
	char log_sep = '/';

	if (!(log_fmt = (char *) malloc(GD_FMT_SIZE))) {
		die("ERROR: Fatal malloc error: util.c get_date_format!");
	}
	...
	switch (set.log_datetime_separator) { ... }   /* three cases */
	switch (set.log_datetime_format)    { ... }   /* six cases, each an snprintf */

	return (log_fmt);
}
```

`spine_log()` frees it again at `util.c:1593`, so there is no leak. But the two inputs, `set.log_datetime_separator` and `set.log_datetime_format`, are read once in `read_config_options()` and never change afterwards, so the whole result is invariant for the life of the process. Every log line pays a `malloc`/`free` pair, two range checks, a three-way switch and a six-way switch to rebuild a string that is always the same 20 bytes.

Suggested: compute the format once after `read_config_options()` and have `get_date_format()` return a pointer to that storage, dropping the per-line allocation. `spine_log()` stops freeing it.

I have not profiled spine, so I am not claiming a measured win. The argument is that the work is provably redundant, the call site is per log line, and the change is small and self-contained. This is the same function that carried the missing-`break` bug fixed in #554, so it has had recent eyes on it.

Two things deliberately not proposed here:

The `strlen(log_fmt) == 0` check at `util.c:1479` is dead, since every path through the switch writes a format. Not worth changing on its own.

`is_debug_device()` is called unconditionally in the `if (is_debug_device(host->id))` / `SPINE_LOG` / `else SPINE_LOG_DEBUG` pattern at 59 sites, before either macro gets to short-circuit on `set.log_level`. That looked like a second easy win, but the function breaks out of its scan at `debug_devices[0] == '\0'`, so with no debug devices configured it is a single array read. Not worth an issue.

PR #576, opened 2026-08-31
fix: unchecked allocations, log newline overflow, and leaked result sets

### LINKED ISSUE #564: Unchecked calloc() results in spine.c dereferenced immediately after allocation
Four `calloc()` results in `spine.c` are used without checking for `NULL`. #542 gave the two in `poller.c` a `die()` on failure; these were never touched.

```c
spine.c:248  php_processes  = (php_t*) calloc(MAX_PHP_SERVERS, sizeof(php_t));
spine.c:254  debug_devices  = calloc(MAX_DEBUG_DEVICES, sizeof(int));
spine.c:555  db_pool_local  = (pool_t *) calloc(set.threads, sizeof(pool_t));
spine.c:563  db_pool_remote = (pool_t *) calloc(set.threads, sizeof(pool_t));
```

Each is dereferenced immediately after:

- `php_processes` is written in the loop on the next line, `php_processes[i].php_state = PHP_BUSY`.
- `debug_devices` is written at `spine.c:539` and `spine.c:545`, and read by `is_debug_device()` at `util.c:417` on every logging call site that uses the debug-device pattern (59 of them across `ping.c`, `poller.c`, `snmp.c` and `util.c`).
- `db_pool_local` and `db_pool_remote` are populated by `db_create_connection_pool()` on the following line.

`set.threads` is operator-controlled, so the pool allocations are the two that can plausibly fail on a constrained host.

`snmp.c:1028` and `util.c:124` already check their results, so this is the remaining set.

Expected: `die()` with the same message shape `poller.c` now uses, rather than dereferencing `NULL`.

### LINKED ISSUE #565: spine_log() can write the newline one byte past flogmessage when the log line fills LOGSIZE
`spine_log()` bounds both of its `strncat()` calls so the formatted message can fill `flogmessage` exactly, then appends a newline unconditionally.

```c
/* util.c:1443 */
char flogmessage[LOGSIZE];      /* LOGSIZE is 65535 */
...
/* util.c:1516-1525 */
if (prefix_len > LOGSIZE - flog_len - 1) {
	prefix_len = LOGSIZE - flog_len - 1;
}

if (ulog_len + prefix_len > LOGSIZE - flog_len - 1) {
	ulog_len = LOGSIZE - flog_len - prefix_len - 1;
}

strncat(flogmessage, logprefix,   prefix_len);
strncat(flogmessage, ulogmessage, ulog_len);
...
/* util.c:1547-1549 */
if (!strstr(flogmessage, "\n")) {
	strcat(flogmessage, "\n");
}
```

After both `strncat()` calls the content can be `flog_len + prefix_len + ulog_len == LOGSIZE - 1` characters, with the terminator at index `LOGSIZE - 1`. The buffer is then full. `strcat()` writes the newline at `LOGSIZE - 1` and its terminator at `LOGSIZE`, one byte past the end of the stack buffer.

Reaching it needs a log line near 64KB, so this is not a routine path, but message text does carry SNMP values and script server output. It is the same class as issue#561 in `php_readpipe()`: capacity was computed without reserving room for the terminator.

Expected: reserve the byte, or use a bounded append that cannot exceed `sizeof(flogmessage) - 1`.

The other eight `strcat()` calls in the tree (`util.c:885-914`) append compile-time literals into `BUFSIZE` buffers that were just `memset()` to zero, total well under 100 bytes. They are safe by construction and are not part of this.

### LINKED ISSUE #566: MYSQL_RES leaked on the NULL-row branch of four util.c settings helpers
Four settings helpers in `util.c` return without freeing the result set when the query returned rows but `mysql_fetch_row()` yields `NULL`.

```c
/* the shape, at util.c:203, 295, 390 and 2224 */
if (result != 0) {
	if (mysql_num_rows(result) > 0) {
		mysql_row = mysql_fetch_row(result);

		if (mysql_row != NULL) {
			retval = strdup(mysql_row[0]);
			db_free_result(result);       /* freed here */
			return retval;
		} else {
			return strdup("");            /* result leaked */
		}
	} else {
		db_free_result(result);           /* freed here */
		return 0;
	}
}
```

The affected functions are `getsetting()` (`util.c:203`), `getpsetting()` (`util.c:295`), `getglobalvariable()` (`util.c:390`), and the Cacti version read at `util.c:2224`. Every other exit in each function frees the result; only the `NULL`-row branch does not.

`mysql_num_rows()` reporting rows and the fetch then returning `NULL` is unusual, so this is a slow leak rather than a fast one, but `getsetting()` is called repeatedly during config load and the version check runs at startup.

Expected: `db_free_result(result)` before returning on that branch, matching the sibling paths.

Note for anyone auditing the same area: all fourteen `mysql_fetch_row()` call sites in the tree are guarded for `NULL`, either by the `while ((row = mysql_fetch_row(result)))` loop condition or by an explicit test. The row handling is fine; only the cleanup on that one branch is missing.

somethingwithproof:
Three defects, one commit each.

**Closes #564 — unchecked `calloc()` in spine.c.** `php_processes`, `debug_devices`, `db_pool_local` and `db_pool_remote` are dereferenced on the statement after allocation. #542 gave the two in poller.c a `die()`; these were missed. All five `calloc()` calls in the file are now guarded.

**Closes #565 — newline written past `flogmessage`.** The two `strncat()` calls are bounded so the message can fill `LOGSIZE` exactly, after which `strcat()` put the newline at `LOGSIZE-1` and its terminator at `LOGSIZE`. The newline is now appended only when a byte is free, and replaces the final character otherwise. Three cmocka cases cover it, with a canary after the buffer so an overflow is observable rather than merely undefined.

**Closes #566 — `MYSQL_RES` leaked on the no-row branch.** `getsetting()`, `getpsetting()`, `getglobalvariable()` and the Cacti version read each returned from the `mysql_fetch_row() == NULL` branch without freeing, while every other exit in those functions frees.

**Verification.** Clean `ubuntu:24.04` container: builds at 4 warnings, matching develop exactly. `make check` passes all three binaries, `make distcheck` passes.

**Not folded into #555.** That PR is ping-specific and currently sits on a stale "merge conflicts" block that only needs lifting; adding unrelated spine.c and util.c changes would force a fresh review of work that is otherwise ready.

The 1.2.x half of these is going onto #574, since all three are present there too.

### DIFF
--- a/spine.c
+++ b/spine.c
@@ -245,13 +245,18 @@ int main(int argc, char *argv[]) {
 	install_spine_signal_handler();
 
 	/* establish php processes and initialize space */
-	php_processes = (php_t*) calloc(MAX_PHP_SERVERS, sizeof(php_t));
+	if (!(php_processes = (php_t*) calloc(MAX_PHP_SERVERS, sizeof(php_t)))) {
+		die("ERROR: Fatal malloc error: spine.c php_processes!");
+	}
+
 	for (i = 0; i < MAX_PHP_SERVERS; i++) {
 		php_processes[i].php_state = PHP_BUSY;
 	}
 
 	/* create the array of debug devices */
-	debug_devices = calloc(MAX_DEBUG_DEVICES, sizeof(int));
+	if (!(debug_devices = calloc(MAX_DEBUG_DEVICES, sizeof(int)))) {
+		die("ERROR: Fatal malloc error: spine.c debug_devices!");
+	}
 
 	/* initialize icmp_avail */
 	set.icmp_avail = TRUE;
@@ -552,7 +557,10 @@ int main(int argc, char *argv[]) {
 	db_connect(LOCAL, &mysql);
 
 	/* setup local connection pool for hosts */
-	db_pool_local = (pool_t *) calloc(set.threads, sizeof(pool_t));
+	if (!(db_pool_local = (pool_t *) calloc(set.threads, sizeof(pool_t)))) {
+		die("ERROR: Fatal malloc error: spine.c db_pool_local!");
+	}
+
 	db_create_connection_pool(LOCAL);
 
 	if (set.poller_id > 1 && set.mode == REMOTE_ONLINE) {
@@ -560,7 +568,10 @@ int main(int argc, char *argv[]) {
 		mode = REMOTE;
 
 		/* setup remote connection pool for hosts */
-		db_pool_remote = (pool_t *) calloc(set.threads, sizeof(pool_t));
+		if (!(db_pool_remote = (pool_t *) calloc(set.threads, sizeof(pool_t)))) {
+			die("ERROR: Fatal malloc error: spine.c db_pool_remote!");
+		}
+
 		db_create_connection_pool(REMOTE);
 	} else {
 		mode = LOCAL;
--- a/tests/unit/test_safety_fixes.c
+++ b/tests/unit/test_safety_fixes.c
@@ -555,8 +555,77 @@ static void test_php_readpipe_bounded(void **state) {
 /* -------------------------------------------------------------------------
  * main
  * ------------------------------------------------------------------------- */
+
+/* ------------------------------------------------------------------------
+ * spine_log(): the newline must not be written past flogmessage
+ *
+ * The strncat() calls above the append are allowed to fill the buffer
+ * exactly, after which strcat() wrote the newline at LOGSIZE-1 and its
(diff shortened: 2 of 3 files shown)


somethingwithproof:
Superseded by #578, which carries the same `spine.c` allocation guards, a larger `util.c` change that also covers #567, and better coverage: five cases against the shipped `spine_log()` rather than the three here, which test a copy of `append_newline()` declared in the test file.

Closing to keep one review rather than two for the same fixes. #564, #565 and #566 are tracked on #578.

PR #577, opened 2026-08-31
test: link the shipped objects instead of copies of the logic

somethingwithproof:
The three unit binaries in `tests/unit/` do not test the code that ships. `test_safety_fixes.c` says so in its own header:

> the functions under test are inlined (or their exact predicate replicated) so this translation unit compiles standalone without linking util.o/ping.o/php.o and dragging in MySQL/SNMP

That keeps them cheap, but a fix can land in `util.c` while the test carries on passing against its private copy. The 39 assertions currently guarantee less than they appear to.

This adds a fourth binary that links the real translation units. `tests/fuzz/stubs.c` already solves the hard part for the fuzz targets by supplying the globals `spine.c` would define, so the same file is reused here. Everything except `spine.c` is linked as built.

**Demonstration.** Reverting `strncopy()` to the issue#447 off-by-one and running the suite:

```
PASS: tests/unit/test_util_strings
PASS: tests/unit/test_build_fixes
PASS: tests/unit/test_safety_fixes
FAIL: tests/unit/test_linked
# TOTAL: 4   # PASS: 3   # FAIL: 1
```

The three existing suites do not notice. Only the linked one does.

Seven cases to start, covering `strncopy()` bounds and termination and `regex_replace()` match/no-match/bad-pattern. The point of the PR is the harness rather than the case count: with this in place, future tests can assert against real behaviour, and the existing binaries can migrate over one at a time.

Verified in a clean `ubuntu:24.04` container: builds at 4 warnings matching develop, `make check` passes 4/4, `make distcheck` passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,40 @@ jobs:
 #            --suppress=toomanyconfigs \
 #            *.c *.h
 
+  # Coverage is reported, not gated.  The number is only meaningful once
+  # something measures it; the threshold conversation comes after that.
+  coverage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Install build dependencies
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y mysql-server libmysqlclient-dev libsnmp-dev \
+            libssl-dev build-essential help2man autoconf automake libtool \
+            dos2unix libcmocka-dev gcovr
+
+      - name: Build instrumented
+        run: |
+          set -euo pipefail
+          ./bootstrap
+          ./configure CFLAGS="-g -O0 --coverage" LDFLAGS="--coverage"
+          make -j"$(nproc)"
+
+      - name: Run the unit tests
+        run: make check
+
+      - name: Report coverage
+        run: |
+          set -euo pipefail
+          # configure leaves instrumented conftest objects behind and gcovr
+          # cannot resolve their sources
+          rm -f ./*conftest*.gcno ./*conftest*.gcda
+          gcovr -r . --exclude 'tests/.*' --txt --print-summary
+
   flawfinder:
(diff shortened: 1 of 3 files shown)


somethingwithproof:
Extended with the coverage work, so this PR now answers "what percentage do we have" as well as fixing the harness.

**The number today is 7 percent, and 0 percent of the shipped code.** Measured with gcov on develop: every one of the eleven production sources reports `never executed`. Only three `.gcda` files are produced against fifteen `.gcno`, and all three belong to test binaries. The 396 covered lines are the test files measuring their own inlined copies.

**Unit tests here take util.c from 0 to 27 percent.** Thirty cases against the linked objects: the string and numeric predicates, the trim family, `read_spine_config()` driven by a real file on disk, every `get_date_format()` format and separator pair, the ICMP classifier across all five verdicts, and `get_namebyhost()` called twice to show the save pointer is no longer shared.

**The integration fixture was already at 38.9 percent and nobody reported it.** Running `tests/snmpv3/scripts/coverage.sh`:

```
spine.c      477   60.2%      poller.c   1154   45.1%
locks.c      167   55.1%      util.c      981   41.6%
keywords.c    27   51.9%      snmp.c      526   36.9%
sql.c        279   49.5%      error.c      86   27.9%
nft_popen.c  127   48.8%      ping.c      666   16.4%
                              php.c       291    0.0%
TOTAL       4808   38.9%
```

That script only printed five of the twelve files, which is why the figure had never been stated. A `coverage` job now runs the instrumented build and prints the summary on every pull request. It reports rather than gates; picking a threshold is worth doing once there is a baseline to argue from.

**Two findings worth keeping.** `php.c` is at 0 percent from both paths, so the script server is entirely unexercised. And `configure` leaves instrumented `conftest` objects behind that make `gcovr` fail outright until they are removed, which is handled in the job.

One test I wrote had to be deleted rather than kept: I assumed `set_option()` applied a value to `set`, and the linked harness failed it immediately. It only records into a file-static override table, so there is nothing observable to assert from outside. That is the harness earning its place on its first run.

TheWitness (approved):

PR #578, opened 2026-08-31
fix: allocation checks, a log buffer overflow, result leaks, and a per-line rebuild

somethingwithproof:
Four small fixes in `spine.c` and `util.c`, all found while auditing classes that #542 and #557 fixed elsewhere in the tree but did not finish.

Based on current develop, after #569 and #570.

**#564** — `php_processes`, `debug_devices`, `db_pool_local` and `db_pool_remote` were dereferenced on the line after `calloc()` without testing the result. `#542` gave the two in `poller.c` a `die()`; these were missed. Same treatment, same message shape. `snmp.c:1028` and `util.c:124` already checked theirs.

**#565** — `spine_log()` bounds both `strncat()` calls so the message may fill `flogmessage` exactly, then appended the newline unconditionally, putting the terminator one byte past a `LOGSIZE` stack buffer. Same class as issue#561. The append is now bounded.

**#566** — four settings helpers returned without `db_free_result()` when the query reported rows but the fetch yielded `NULL`. Every other exit in each function frees it.

**#567** — `get_date_format()` did a `malloc`/`free` pair, two range checks and two switches on every log line to rebuild a string that only depends on two settings read once at config load. It is now built in `read_config_options()`, while spine is still single threaded, and `get_date_format()` returns that storage. `spine_log()` no longer frees it. No behaviour change; the format for every code is unchanged, which the existing tests in `test_safety_fixes.c` already assert.

**Verification.** Clean `ubuntu:24.04` container: builds at 4 warnings, matching develop, and `make check` passes all four suites. The new test is wired into the `check_PROGRAMS` list #570 added.

`tests/unit/test_log_newline_bound.c` is new and pins #565 down. It fails against the old `strcat()`: built with ASan and the pre-fix predicate, `test_full_buffer_is_left_alone` reports a stack-buffer-overflow; with the fix all five pass.

**Two things deliberately not here.** All fourteen `mysql_fetch_row()` call sites are already guarded, by the loop condition or an explicit test, so there was nothing to fix beyond the leak in #566. And the eight `strcat()` calls at `util.c:885-914` append compile-time literals into `BUFSIZE` buffers that were just zeroed, well under 100 bytes total; they are safe by construction and left alone.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,10 @@ The Cacti Group | spine
 -issue#552: Terminate die() output so consecutive fatal messages no longer run together
 -issue#561: Reserve room for the terminator in php_readpipe() so a full script server result cannot write past result_string
 -issue#562: Escalate PHP script server shutdown to SIGKILL after a bounded grace period so a stuck child is not orphaned
+-issue#564: Check the calloc() results in spine.c so an allocation failure dies instead of dereferencing NULL
+-issue#565: Bound the newline appended by spine_log() so a full log line cannot write past flogmessage
+-issue#566: Free the result set on the NULL-row branch of the util.c settings helpers
+-issue#567: Build the log timestamp format once at config load instead of rebuilding it on every log line
 -issue: Correct signed and unsigned printf format specifiers in poller.c, free session.localname on the unknown-version return in snmp.c, and quote shell variables in the build scripts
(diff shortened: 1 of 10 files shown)


TheWitness (dismissed):

TheWitness:
Merge conflicts.

somethingwithproof:
Conflicts are cleared. GitHub reports this `MERGEABLE` against current develop, and #576 is closed as a duplicate of this one, so the overlap on `spine.c` and `util.c` is gone.

somethingwithproof:

PR #584, opened 2026-09-02
test: cover the poller result predicates against the shipped objects

somethingwithproof:
`poller.c` is the most frequently changed file in the tree, 28 of the last 100 commits, and `validate_result()` is where it decides whether a polled value is stored at all. Neither it nor `is_multipart_output()` had any coverage. A mistake in either silently discards data or accepts a malformed multi-value string, and nothing fails.

Six cases added to `test_linked`, so they run against the real `poller.o` rather than a copy of the logic:

- numeric forms: integer, negative, decimal, zero
- rejection: `NULL`, free text, empty string
- multi-field output, single and several pairs
- the space-to-delimiter balance, which is what separates `in:100 out:200` from prose that merely contains a colon such as `time is 12:30 today`
- the `trim()` contract

## On the trim case
It documents an asymmetry worth knowing about. `rtrim()` writes a NUL over the trailing run, so it edits the caller's buffer. `ltrim()` only advances a pointer, so the leading run survives in the buffer the caller keeps. `validate_result()` therefore mutates what it is handed, but not into the string the predicate actually saw. The trim set is also wider than whitespace: it includes quotes and a backslash.

I wrote that test asserting the tidier behaviour first, `"  field:1  "` becoming `"field:1"`. It failed, and the assertions here describe what the code does.

## Verification
Built and run in an Ubuntu 24.04 container matching `ci.yml`'s dependency list:

```
PASS: tests/unit/test_util_strings
PASS: tests/unit/test_build_fixes
PASS: tests/unit/test_safety_fixes
PASS: tests/unit/test_linked
# TOTAL: 4   # PASS: 4   # FAIL: 0

test_linked: 43 tests run, 43 passed
```

No build change is needed: `Makefile.am` already links `poller.c` into `test_linked`.

Follows the pattern established by #563, which found the suite was certifying copies rather than shipped code.

### DIFF
--- a/tests/unit/test_linked.c
+++ b/tests/unit/test_linked.c
@@ -457,6 +457,91 @@ static void test_is_debug_device_matches_only_listed_ids(void **state) {
 	debug_devices = saved;
 }
 
+
+/* poller.c decides here whether a polled value is stored at all. A result that
+ * validate_result() rejects is discarded, so a mistake in either predicate
+ * silently drops data or accepts a malformed multi-value string. poller.c is
+ * the most frequently changed file in the tree and had no coverage of either.
+ */
+static void test_validate_result_accepts_numeric_forms(void **state) {
+	char integer[]  = "42";
+	char negative[] = "-17";
+	char decimal[]  = "3.14159";
+	char zero[]     = "0";
+	(void) state;
+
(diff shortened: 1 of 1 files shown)


somethingwithproof:

PR #585, opened 2026-09-02
fix: size the escape destination from the caller's buffer

### LINKED ISSUE #583: Poller results of 1024 bytes or more are silently truncated to 1022
`926d5b6` correctly replaced an unescaped interpolation of poller results with `db_escape()`, but sized the destination at half what it needs, so any result of 1024 bytes or more is silently cut to 1022 characters.

`poller.c:1885-1889`:

```c
char escaped_result[DBL_BUFSIZE];        /* 2048 */
db_escape(&mysqlt, escaped_result, sizeof(escaped_result), poller_items[i].result);
```

`poller_items[i].result` is `char[RESULTS_BUFFER]`, which defaults to 2048 and is user-tunable through `--with-results-buffer`. Escaping can double the input, so `db_escape` (`sql.c:586-603`) trims:

```c
max_escaped_input_size = (strlen(input) * 2) + 1;
trim_limit = (max_size < DBL_BUFSIZE) ? max_size : DBL_BUFSIZE;
if (max_escaped_input_size > max_size) {
    snprintf(input_trimmed, (trim_limit / 2) - 1, "%s", input);
```

With a 2048-byte destination, `(2048 / 2) - 1` is 1023, so at most 1022 characters survive:

```
input 1023 bytes -> 1023 survive
input 1024 bytes -> 1022 survive   truncated
input 2047 bytes -> 1022 survive   truncated
```

Raising `--with-results-buffer` does not help, because `db_escape` also caps its own `input_trimmed[DBL_BUFSIZE]` regardless of what the caller passes.

Failure scenario: a multi-output data source whose `field:value` list exceeds 1023 bytes is cut mid-token, so Cacti's `poller_output` parser sees a malformed trailing pair and that data source gets a bad or missing RRD update every cycle. `RESULTS_BUFFER` exists precisely to allow results of this size.

The escaping itself is correct and should stay. Only the sizing is wrong: `escaped_result` needs `RESULTS_BUFFER * 2 + 1`, and `db_escape`'s internal buffer needs to scale with `max_size` rather than being fixed at `DBL_BUFSIZE`.

The pull request's own test plan left "Verify poller output results are correctly escaped in the database" unchecked.

somethingwithproof:
Fixes #583, and the tests written for it turned up something worse than the reported truncation.

## The reported bug
`db_escape()` staged its input through a fixed `DBL_BUFSIZE` buffer before escaping:

```c
trim_limit = (max_size < DBL_BUFSIZE) ? max_size : DBL_BUFSIZE;
if (max_escaped_input_size > max_size) {
    snprintf(input_trimmed, (trim_limit / 2) - 1, "%s", input);
```

Every caller was therefore capped at 1022 input bytes regardless of the destination it passed. A poller result of 1024 bytes or more was silently cut, which matters because `RESULTS_BUFFER` defaults to 2048 and is tunable through `--with-results-buffer` precisely to carry multi-value output. A cut result loses its trailing `field:value` pair, so that data source gets a bad RRD update every cycle.

## What the tests found as well
The same arithmetic is wrong for a small destination. With `max_size` of 1, `(trim_limit / 2) - 1` is `-1`, which `snprintf` takes as an unbounded size, and the escaped result is then written into a destination far smaller than it.

Building the new tests against the **old** implementation does not just fail an assertion:

```
[  FAILED  ] test_db_escape_truncates_into_a_small_destination
[ RUN      ] test_db_escape_handles_a_degenerate_destination
*** buffer overflow detected ***: terminated
```

No in-tree caller passes a destination that small today, so this is a latent overflow rather than a live one, but it is the reason the function needed restructuring rather than a wider staging buffer.

## The change
Derive the limit from the destination and escape straight from the caller's input, dropping the staging buffer entirely. Escaping can double every byte and the result is NUL terminated, so at most `(max_size - 1) / 2` input bytes fit.

`poller.c` sizes `escaped_result` at `(RESULTS_BUFFER * 2) + 1` for its `RESULTS_BUFFER` source, which is what the doubling requires. `escaped_rrd_name` is left alone; its source is `char[30]`.

## Tests
Six cases in `test_linked`, so they run against the shipped `sql.o`:

- metacharacter escaping, quote and backslash
- a `NULL` input leaves the destination untouched
- a full `RESULTS_BUFFER` source round trips with nothing dropped
- the 1022, 1023, 1024, 1100, 2047 boundary the old staging buffer cut at
- truncation into a destination too small for the escaped form, still terminated
- a degenerate `max_size` of 0 and 1

Verified in an Ubuntu 24.04 container matching `ci.yml`: 43 of 43 tests pass with the fix, and the boundary and degenerate cases fail without it.

### DIFF
--- a/poller.c
+++ b/poller.c
@@ -1882,7 +1882,9 @@ void poll_host(int device_counter, int host_id, int host_thread, int host_thread
 
 		i = 0;
 		while (i < rows_processed) {
-			char escaped_result[DBL_BUFSIZE];
+			/* escaping can double every byte, so the destination has to be
+			 * twice the source plus the terminator */
+			char escaped_result[(RESULTS_BUFFER * 2) + 1];
 			char escaped_rrd_name[DBL_BUFSIZE];
 
 			db_escape(&mysqlt, escaped_result, sizeof(escaped_result), poller_items[i].result);
--- a/sql.c
+++ b/sql.c
@@ -584,22 +584,32 @@ int append_hostrange(char *obuf, const char *colname) {
  *
  */
 void db_escape(MYSQL *mysql, char *output, int max_size, const char *input) {
-	char input_trimmed[DBL_BUFSIZE];
-	int  max_escaped_input_size;
-	int  trim_limit;
+	size_t input_len;
+	size_t max_input;
 
-	if (input == NULL) return;
+	if (input == NULL || output == NULL) return;
 
-	max_escaped_input_size = (strlen(input) * 2) + 1;
-	trim_limit = (max_size < DBL_BUFSIZE) ? max_size : DBL_BUFSIZE;
+	if (max_size <= 1) {
+		if (max_size == 1) {
+			*output = '\0';
+		}
 
-	if (max_escaped_input_size > max_size) {
-		snprintf(input_trimmed, (trim_limit / 2) - 1, "%s", input);
-	} else {
-		snprintf(input_trimmed, trim_limit, "%s", input);
+		return;
 	}
 
-	mysql_real_escape_string(mysql, output, input_trimmed, strlen(input_trimmed));
+	/* Escaping can double every byte and the result is NUL terminated, so at
+	 * most (max_size - 1) / 2 input bytes fit. The previous implementation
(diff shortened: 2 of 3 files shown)


somethingwithproof:

Issue #373, opened 2025-09-19
Spine cannot generate SNMPv3 Ku for SHA-384/SHA-512


Kveri:
**Describe the bug**
I'm running default spine 1.2.30 / cacti 1.2.30. I face few problems with snmpv3 and SHA above 256.

1. while using php-snmp lib, it doesn't work even in device detail (in browser), the top part of the page where SNMP info is displayed doesn't display it.
2. after disabling php-snmp lib and relying on net-snmp, then in-browser SNMP details are displayed, even with SHA-512 as auth protocol (AES-256 as priv protocol). However, spine can't poll such a device.
3. As a side bug: In Device settings the SNMP Auth Protocol selection - SHA-392 is invalid, it should be SHA-384

**To Reproduce**
1. Install vanilla cacti + spine 1.2.30
2. add device with AES-256 / SHA-512 SNMPv3 security
3. observe the issue

**Expected behavior**
It should work.

I troubleshot this by one-shot running spine:
`spine -H 28 -R -S -P -M -t 1 -V 100`

I get these errors:
```
Total[0.3906] SNMP: Error generating SNMPv3 Ku from authentication pass phrase.
Total[0.3907] SNMP: Using privacy protocol(len): AES256(256)
Total[0.3907] SNMP: Error generating SNMPv3 Ku from privacy pass phrase.
```

With SHA-384 (broken 392) and SHA-512 this is the error. With SHA-256 there is no error and polling works.

Troubleshooting further I added some debug outputs to spine snmp.c, I verified that session.securityAuthProto is valid (the respective OIDs are valid), but generate_Ku simply returns "Error: unknown authtype". Digging deeper into net-snmp scapi.c sc_get_authtype() and subsequenty into sc_find_auth_alg_byoid() which just looks into _auth_alg_info array where this element exists:
```
#ifdef HAVE_EVP_SHA384
    { NETSNMP_USMAUTH_HMAC256SHA384, "usmHMAC256SHA384AuthProtocol",
      usmHMAC256SHA384AuthProtocol, OID_LENGTH(usmHMAC256SHA384AuthProtocol),
      BYTESIZE(SNMP_TRANS_AUTHLEN_HMAC256SHA384), USM_HMAC256SHA384_AUTH_LEN },
    { NETSNMP_USMAUTH_HMAC384SHA512, "usmHMAC384SHA512AuthProtocol",
[...]

PR #586, opened 2026-09-02
fix: restore the SNMPv3 noAuthNoPriv and authNoPriv security levels

### LINKED ISSUE #582: SNMPv3 authNoPriv and noAuthNoPriv are non-functional since #373
Two of SNMPv3's three security levels do not work in spine. Both are offered by Cacti's UI and both work under `cmd.php`, so a device that polls fine with the PHP poller fails with spine.

Introduced by `fce0b42` (Fix #373, SHA-512 auth).

## authNoPriv authenticates with an empty key

`snmp.c:244-254` sets `securityLevel = SNMP_SEC_LEVEL_AUTHNOPRIV` when privacy is `[None]` or the passphrase is empty, but never derives the authentication key. The only `generate_Ku()` calls are at `snmp.c:312` and `snmp.c:350`, both inside the privacy branch. `securityAuthKeyLen` therefore stays 0.

Before `fce0b42`, `generate_Ku()` ran unconditionally whenever `snmp_password` was non-empty.

Reproduced against net-snmp 5.9.5.2 and a live `snmpd`, replicating both code paths:

```
current develop      securityAuthKeyLen=0   -> FAILED: USM authentication failure
pre-fce0b42          securityAuthKeyLen=20  -> SUCCESS: STRING: spine-audit-agent
```

The device is marked down and the log blames the credentials, which points operators at their SNMP configuration rather than at spine.

## noAuthNoPriv is refused before the session opens

`snmp.c:231` calls `usm_lookup_auth_type(snmp_auth_protocol)` and returns 0 on any non-positive result, after logging `Error auth protocol ... is invalid`.

Cacti stores `[None]` for "no auth protocol" (`global_arrays.php:657`), and `usm_lookup_auth_type("[None]")` returns -1 in net-snmp 5.9.5.2. So the session is never created.

Before `fce0b42` the if/else chain simply matched nothing, left `securityAuthProto` NULL, and fell through to `SNMP_SEC_LEVEL_NOAUTH`, which net-snmp accepts.

## Why no test caught it

The integration fixture uses SHA-256 with AES and a privacy passphrase, which is authPriv. That is the one v3 branch still working, so neither level above is covered.

## Suggested fix
Treat `[None]` and an empty protocol as "no authentication" before the lookup, and derive the authentication key whenever the level is authNoPriv or authPriv rather than only in the privacy branch.

somethingwithproof:
Fixes #582.

Cacti stores the literal `[None]` when no SNMPv3 protocol is selected, and an empty string for an absent passphrase. Since #373 spine treated both as errors, so two of the three security levels stopped working while `cmd.php` kept accepting them. A device that polls fine under the PHP poller fails under spine.

Measured against real net-snmp, replicating the shipped predicates:

```
level          | old: session opens? | old: derives auth key?
noAuthNoPriv   | NO  <-- broken      | n/a
authNoPriv     | yes                 | NO  <-- broken
authPriv       | yes                 | yes
```

**noAuthNoPriv** — `usm_lookup_auth_type("[None]")` returns `-1`, and `snmp.c` returned 0 from the session setup on any non-positive result, so the session was never opened.

**authNoPriv** — the authentication key was derived only on the privacy path, so the session authenticated with `securityAuthKeyLen` still 0 and the agent answered with a USM authentication failure. That points an operator at their credentials rather than at spine.

## The change
Decide the level first, from the four values Cacti supplies, and only require a recognised protocol when authentication is actually in use. Derive the key whenever the level authenticates rather than only when it also encrypts.

The decision is extracted as `spine_snmpv3_security_level()`, which is pure and therefore testable without a live session. That is deliberate: the logic that was wrong is now the part that has coverage.

## Tests
Five cases in `test_linked`, against the shipped `snmp.o`:

- `[None]`, empty and `NULL` are unset; `none` and `[None] ` are not the sentinel
- noAuthNoPriv for no protocol, and for a protocol without a password
- authNoPriv for auth without privacy, including a privacy protocol whose passphrase is missing
- authPriv when both are set
- privacy selected without authentication does not raise the level, since SNMPv3 has no privNoAuth

42 of 42 pass in an Ubuntu 24.04 container matching `ci.yml`, and `snmp.c` compiles warning-clean under `--enable-warnings`.

## Not covered here
The integration fixture still exercises authPriv only. Adding noAuthNoPriv and authNoPriv devices to it would have caught this and is worth doing separately.

### DIFF
--- a/snmp.c
+++ b/snmp.c
@@ -41,6 +41,62 @@
 
 #define OIDSIZE(p) (sizeof(p)/sizeof(oid))
 
+/*! \fn int spine_snmpv3_value_is_set(const char *value)
+ *  \brief Whether a Cacti-supplied SNMPv3 field selects anything.
+ *
+ *  Cacti stores the literal "[None]" when the user picks no protocol, and an
+ *  empty string when a passphrase is absent. Both mean "not selected".
+ */
+int spine_snmpv3_value_is_set(const char *value) {
+	if (value == NULL) {
+		return FALSE;
+	}
+
+	if (value[0] == '\0') {
+		return FALSE;
+	}
+
+	if (strcmp(value, "[None]") == 0) {
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+/*! \fn int spine_snmpv3_security_level(...)
+ *  \brief Pick the SNMPv3 security level from the values Cacti stores.
+ *
+ *  Authentication needs both a protocol and a password; privacy additionally
+ *  needs a privacy protocol and passphrase, and is only meaningful on top of
+ *  authentication. Anything not selected leaves the level lower rather than
+ *  making the device an error, which is what Cacti's own poller does.
+ *
+ *  \return SNMP_SEC_LEVEL_NOAUTH, SNMP_SEC_LEVEL_AUTHNOPRIV or
+ *          SNMP_SEC_LEVEL_AUTHPRIV
+ */
+int spine_snmpv3_security_level(const char *auth_protocol, const char *auth_password,
+		const char *priv_protocol, const char *priv_passphrase) {
+	int authenticates;
+	int encrypts;
+
+	authenticates = spine_snmpv3_value_is_set(auth_protocol) &&
+		spine_snmpv3_value_is_set(auth_password);
+
+	encrypts = authenticates &&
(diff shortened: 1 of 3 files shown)


somethingwithproof:

PR #589, opened 2026-09-02
Build poll_host queries once instead of once per poller type

### LINKED ISSUE #587: poll_host() duplicates its query construction across the local and remote paths
`poll_host()` in `poller.c` is 1,926 lines, more than four times the next largest function in the tree (`read_config_options()` at 419). `poller.c` is also the most frequently changed file, 28 of the last 100 commits, so this is where regressions are most likely and hardest to see.

Two blocks inside it, at roughly `poller.c:306` and `poller.c:448`, build the same set of queries and differ only in one WHERE clause:

```
block 1:  AND deleted = ''
block 2:  AND poller_id=%i
```

Everything else is duplicated verbatim, including the fifteen-line `host` SELECT that lists every column the uptime check reads. Measured at roughly 100 lines of duplication.

The cost is not the line count. A column added to one copy and not the other produces a mismatch between the local and remote poller paths that nothing would catch, because none of the query construction in `poll_host()` is reachable from a test.

## Proposal
Extract the query construction into a helper that takes the host id and whichever discriminator distinguishes the two paths, returning the built strings. That removes the duplication and, more usefully, gives the query text a seam a unit test can reach.

Not a behaviour change. The extracted helper should produce byte-identical SQL for both existing paths, and that equivalence is what the tests should assert.

### LINKED ISSUE #590: Remote pollers ignore dbonupdate and emit the deprecated VALUES() upsert form
## What happens

`poll_host()` builds its queries twice, once under `if (set.poller_id == 0)` and once in the `else`. The main-poller copy adapts the `poller_output` insert suffix to the database version. The remote copy does not:

```c
/* poller.c:386, main poller */
if (set.dbonupdate == 0) {
    snprintf(posuffix, BUFSIZE, " ON DUPLICATE KEY UPDATE output=VALUES(output)");
} else {
    snprintf(posuffix, BUFSIZE, " AS rs ON DUPLICATE KEY UPDATE output=rs.output");
}

/* the remote branch, same buffer, no version check */
snprintf(posuffix, BUFSIZE, " ON DUPLICATE KEY UPDATE output=VALUES(output)");
```

`set.dbonupdate` is set in `util.c:375` from the server banner: 0 for MariaDB, 1 for MySQL 8.x. So a remote poller against MySQL 8 keeps using `VALUES(output)`, which MySQL deprecated in 8.0.20 in favour of the row-alias form. It still executes today, so this is not an outage; it emits deprecation warnings now and is scheduled to stop working.

## Proof

Compiling both branch bodies verbatim as a standalone generator and printing `posuffix` for each input vector:

```
MAIN   ports=1 onupd=0  ->  ON DUPLICATE KEY UPDATE output=VALUES(output)
MAIN   ports=1 onupd=1  ->  AS rs ON DUPLICATE KEY UPDATE output=rs.output
REMOTE ports=1 onupd=0  ->  ON DUPLICATE KEY UPDATE output=VALUES(output)
REMOTE ports=1 onupd=1  ->  ON DUPLICATE KEY UPDATE output=VALUES(output)   <-- should match MAIN
```

## Two related loose ends in the same area

`read_config_options()` in `util.c:366` decides whether it is talking to a remote database and then never uses the answer:

```c
if (set.poller_id > 1 && set.mode == REMOTE_ONLINE) {
    db_connect(REMOTE, &mysqlr);
    mode = REMOTE;
} else {
    mode = LOCAL;
}
/* mode is never read again; the version query below is hardcoded LOCAL */
if ((res = getglobalvariable(&mysql, LOCAL, "version")) != 0) {
```

And `config_t` carries `rdbversion` and `rdbonupdate` (`spine.h:434-435`), the remote mirrors of `dbversion`/`dbonupdate`. Neither is ever read or written anywhere in the tree. Together those read as a remote-side version detection that was declared and never wired up.

## Why this is worth more than the one-line fix

Nothing in this construction was reachable from a test, and the two copies had already drifted. A column added to one copy and not the other would land the same way. The fix is cheap; the duplication that hid it is #587.

somethingwithproof:
Closes #587. Closes #590.

`poll_host()` is the largest function in the tree by a factor of eight. It built the same six queries twice, once under `if (set.poller_id == 0)` and once in the `else`, 141 and 142 lines differing by 33. None of it was reachable from a test.

The copies had already drifted. The main poller adapts the `poller_output` upsert to the server version; the remote copy never picked that up, so a remote poller against MySQL 8 still emits `VALUES(output)`, deprecated in 8.0.20. That is #590, and it is the argument for the refactor rather than a separate cleanup.

## Three commits

**1. Collapse the two branches.** Both bodies compile standalone once eight inputs are declared, so the whole construction could be lifted into a generator and its output compared before and after. Across every vector it branches on, `total_snmp_ports` 1 and 2, `dbonupdate` 0 and 1, main and remote, the 72 strings are identical except four lines where `query1` on a remote poller now spells its tail `poller_id = N` rather than `poller_id=N`. The other five queries already used spaces. 286 lines become 153.

**2. Fix #590.** After the collapse it is the removal of one guard clause. Two lines of the capture change, both `posuffix` on a remote poller with `dbonupdate=1`.

**3. Extract the construction into `poll_host_build_queries()`.** It takes `host_id`, `regex_col` and `limits` as arguments, reads five documented settings, and fills a `poll_host_queries_t`. Nothing else moves.

## What that buys, concretely

`poll_host()` goes from **1,927 lines to 1,620**. That is a first cut, not a decomposition; the three big blocks left inside it are 697, 362 and 293 lines and each needs the same treatment.

The part that matters more is that the construction is now callable, so the capture is a test that runs rather than a file with instructions attached:

```
[  ERROR   ] --- ERROR: query construction diverged from the golden capture at line 3
[  FAILED  ] test_build_queries_matches_the_golden_capture
```

That is the suite failing against a one-character edit to the fixture. `make check` now covers this code.

## Tests

Twelve cases against the shipped `poller.o`, up from none:

- the two scope helpers, on both poller types, plus a degenerate buffer and a `NULL` destination
- the main poller filters by `deleted` and carries no `poller_id` anywhere
- a remote poller filters by owner on `query1`, `query5`, `query9` and `query10`, and the host row is filtered by `deleted` on both, never by owner
- `ORDER BY snmp_port` appears only for multiple ports
- `dbonupdate` applies on both poller types, which is #590 pinned
- the cached lengths match the strings the result loop appends
- every buffer is filled
- the golden diff above

49 of 49 pass. Clean Ubuntu 24.04 rebuild matching `ci.yml` carries the same four warnings as develop.

## Note for review

`query1` was being reused near the end of `poll_host()` for an unrelated `UPDATE host SET polling_time`. It is now `q.query1`, so that reuse is still there and is now more visible than before. Giving it its own buffer is a one-line change I left out of this PR.

### DIFF
--- a/poller.c
+++ b/poller.c
@@ -222,6 +222,8 @@ void poll_host(int device_counter, int host_id, int host_thread, int host_thread
 	char temp_poll_result[BUFSIZE];
 	char temp_arg1[BUFSIZE];
 	char limits[SMALL_BUFSIZE];
+	char item_scope[SMALL_BUFSIZE];
+	char owner_scope[SMALL_BUFSIZE];
 
 	int  last_snmp_version = 0;
 	int  last_snmp_port    = 0;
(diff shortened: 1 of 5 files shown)


somethingwithproof:

PR #591, opened 2026-09-03
Restore the close-on-exec and bounded reap removed by #542

### LINKED ISSUE #581: PR #542 reverted the FD_CLOEXEC and bounded-reap hardening added by #557
PR #542 (`e8a4b9a`) landed three commits after PR #557 (`0c481bc`) and removed the process hardening #557 had just added. The #542 body does not mention the removal. #542 has the lower number, so its branch predates #557 and the base appears to have been folded in rather than rebased.

`git grep CLOEXEC` over `develop` now returns nothing.

## What was removed

**File descriptors no longer close on exec.** `0c481bc` added `set_cloexec()`/`set_pipe_cloexec()` in `nft_popen.c` and the equivalents in `php.c`; `e8a4b9a` deleted all of them (17 references).

Two consequences:

- `pipe(pdes)` in `nft_popen()` runs before `pthread_mutex_lock(&ListMutex)`, and the entry joins `PidList` only after `posix_spawn` returns. A thread spawning inside that window produces a child holding another thread's pipe write end. The first thread's script exits without it ever seeing EOF, so it blocks to `script_timeout` and records `U` for a data source that answered.
- The PHP script-server pipes are never in `PidList` at all, so every script child inherits `cacti2php_pdes[1]`. `php_close()` signals the server by closing that descriptor, so a child holding a duplicate keeps the server alive past shutdown and script servers accumulate across restarts.

**The bounded child reap became an unbounded block.** `0c481bc` had `reap_child_bounded()` using `WNOHANG` with a SIGKILL escalation. `nft_popen.c:359` on develop is now:

```c
do { pid = waitpid(cur->pid, &pstat, 0);
} while (pid == -1 && errno == EINTR);
```

`exec_poll()` only sends `SIGKILL` on the `select()` timeout branch; on the normal branch it reads one line and calls `nft_pclose()` directly. A script that prints its value and then exits slowly, or that ignores SIGPIPE, pins the poller thread with no timeout, across cycles. The thread holds its `available_scripts` token throughout, so enough wedged scripts starve every other poller thread.

## Why this looks accidental rather than intended

The same commit **kept** the equivalent bounded reap in `php.c` as `php_terminate_and_reap()`, with a comment stating the delay is load-bearing. Keeping it in one file and deleting it in the sibling is the signature of a stale-branch merge.

The commit message also mis-describes the code it removed: it says the reap delay "was compiled out on `SOLAR_THREAD` builds", but the removed code reads `#ifndef SOLAR_THREAD usleep(50000) #else sleep(1) #endif`. Neither branch is compiled out.

## Suggested fix
Restore both pieces from `0c481bc`, and set `FD_CLOEXEC` on the pipe before it is exposed to another thread rather than after.

somethingwithproof:
Closes #581.

PR #542 landed three commits after #557 and removed the process hardening #557 had just added. #542 has the lower number, so its branch predates #557 and the base appears to have been folded in rather than rebased. `git grep CLOEXEC` over develop currently returns nothing.

## What was lost

**Pipes are inherited across exec.** `nft_popen()` creates its pipe before taking `ListMutex`, and the entry joins `PidList` only after `posix_spawn` returns. A thread spawning inside that window produces a child holding another thread's pipe write end. The first thread never sees EOF, so it blocks to `script_timeout` and records `U` for a data source that answered. The script-server pipes are worse: they are never in `PidList` at all, so every script child inherits `cacti2php_pdes[1]`, and since `php_close()` signals the server by closing that descriptor, a child holding a duplicate keeps the server alive past shutdown.

**The bounded reap became an unbounded block.** `exec_poll()` only sends `SIGKILL` on the `select()` timeout branch; the normal branch reads one line and calls `nft_pclose()` directly. A script that prints its value and then exits slowly pins the poller thread with no timeout, across cycles, holding its `available_scripts` token throughout.

## What this restores, and what it does not

`php.c` kept its bounded reap through #542 as `php_terminate_and_reap()`, and that version is better than the one that was removed, so it is untouched. Only its pipes needed the flag. `nft_popen.c` needed both halves.

The two files had separate copies of the same close-on-exec helper. There is now one, `spine_open_pipe_cloexec()`, used by both.

Also restored: the `close(cacti2php_pdes[...])` pair on the second pipe's failure path, dropped by the same commit, which leaked two descriptors.

Kept from #542: the `free(cur)` additions on two `nft_popen()` error paths, and the `RESULTS_BUFFER` bounds work in `php.c`. This reverts nothing that commit got right.

`pipe2(pdes, O_CLOEXEC)` would close the window between `pipe()` and `fcntl()` atomically, but it needs `_GNU_SOURCE` on glibc and spine defines no feature macro. Adding one is a portability decision that does not belong in a restore, so the `fcntl()` pair stays and the comment says why.

## Tests

The previous guard was `tests/regression/test_child_process_safety.sh`, which grepped the source for the strings it wanted and was deleted alongside the code it protected. Six cmocka cases replace it, against the shipped `nft_popen.o`:

- the flag is set on both ends
- the pipe still works as a pipe
- the descriptor does not survive an `exec`, checked by having a child look for it in `/proc/self/fd`
- a live child returns "still running" rather than blocking
- an exited child is collected with its status
- a child someone else reaped reports success with no status

Reverting just the `fcntl()` calls, which is what #542 did, fails two of them:

```
[  ERROR   ] --- (flags & FD_CLOEXEC) != 0
[  FAILED  ] test_cloexec_is_set_on_both_pipe_ends
[  ERROR   ] --- 0 != 0x1
[  FAILED  ] test_pipe_is_not_inherited_across_exec
```

The second is the interesting one: `0` is the child reporting it found the descriptor.

43 of 43 pass with the fix, `make check` is green on all four binaries, and the build carries the same four warnings as develop.

### DIFF
--- a/nft_popen.c
+++ b/nft_popen.c
@@ -87,6 +87,8 @@
 #include "common.h"
 #include "spine.h"
 #include <spawn.h>
+#include <fcntl.h>
+#include <sys/wait.h>
 
 /* An instance of this struct is created for each popen() fd. */
 static struct pid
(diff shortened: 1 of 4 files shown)


somethingwithproof:

PR #592, opened 2026-09-03
Replace the 'p += snprintf' idiom with a bounded append

### LINKED ISSUE #588: The 'p += snprintf' idiom moves the pointer past the buffer on truncation
The bounded-append idiom appears at 53 sites, 47 in `util.c` and 6 in `spine.c`:

```c
remaining = SIZE - (p - buf);
p += snprintf(p, remaining, "...", ...);
```

`snprintf` returns the length it *would* have written, not what it wrote, so on the first truncation the pointer moves past the end of the buffer. Demonstrated with a 32-byte buffer:

```
buffer size      : 32
pointer advanced : 40      <-- past the end
next 'remaining' : -8  ->  as size_t: 18446744073709551608
```

The next call then receives an effectively unbounded size and a destination already out of bounds.

`util.c:829-850` is the worst shape: it passes a fixed `BUFSIZE` to every call rather than a decreasing remainder, so the bound never shrinks at all, and line 849 interpolates `set.host_id_list`, which `--hostlist` can fill with up to 65534 bytes.

On the stock schema `poller_push_data_to_main()` lands near 1.75 MB against a 2.048 MB buffer, so this does not overflow today. The margin is about 15 percent and the guard that is supposed to enforce it does nothing.

## Proposal
A single helper that owns the arithmetic and reports truncation rather than walking off the end:

```c
if (!spine_appendf(&p, &remaining, "...", ...)) goto full;
```

This removes no lines. The value is that 53 sites stop re-implementing a bound that is currently wrong, and a reviewer can see correctness at the call site.

somethingwithproof:
Closes #588.

```c
remaining = SIZE - (p - buf);
p += snprintf(p, remaining, "...", ...);
```

`snprintf` returns the length it *would* have written, so the first truncation moves the cursor past the end of the buffer. The next `remaining` is then negative, and as a `size_t` it is effectively unbounded, at a destination that is already out of bounds.

The idiom appears at 53 sites, 47 in `util.c` and 6 in `spine.c`.

## The one that is already wrong

`util.c:829-851` passed a fixed `BUFSIZE` to every call rather than a shrinking remainder, so the bound never moved at all:

```c
sqlp += snprintf(sqlp, BUFSIZE, "SELECT SQL_NO_CACHE action FROM poller_item");
sqlp += snprintf(sqlp, BUFSIZE, " WHERE action=%d", POLLER_ACTION_PHP_SCRIPT_SERVER);
sqlp += snprintf(sqlp, BUFSIZE, " AND host_id IN(%s)", set.host_id_list);
```

`--hostlist` can fill `set.host_id_list` with up to 65534 bytes. The third call writes at most 1023 of them, then advances `sqlp` by about 65550, and the query that reaches `db_query()` is the truncated prefix. The buffer is `HUGE_BUFSIZE`, so this is malformed SQL rather than a stack overflow, but nothing here is doing what it reads as doing.

`poller_push_data_to_main()` is where the margin matters: on the stock schema it lands near 1.75 MB against a 2.048 MB buffer, about 15 percent, with a guard that does not work.

## The change

`spine_appendf()` owns the arithmetic, advances by what was actually written, and returns FALSE on truncation. On truncation it leaves the cursor on the terminator, so the buffer stays a valid string and every later append fails there rather than running past the end.

All 53 sites are converted. **Each recompute line is kept**:

```c
remaining = HUGE_BUFSIZE - (sqlp - sqlbuf);
spine_appendf(&sqlp, &remaining, "%s, ", row[0]);
```

`spine_appendf()` maintains `remaining` itself, so the recompute is redundant. Dropping it would mean finding the right initialisation point for every cursor in a function that resets `sqlp` inside loops. Keeping it makes the conversion purely local: no buffer depends on being initialised somewhere new, and the recompute still self-heals. Removing them is a separate, later change.

Two `int remaining` declarations became `size_t`.

## Tests

Seven cases against the shipped `util.o`, each using a struct with a `0x7e` canary after the buffer so an overflow is observable rather than merely undefined:

- appends and advances by what was written
- successive appends accumulate
- truncation returns FALSE, leaves the cursor inside the buffer on the terminator, and sets `remaining` to 1
- a second append after truncation writes nothing, anywhere
- NULL cursor, NULL target and NULL remainder are refused
- an exhausted buffer is refused without touching it
- `test_old_idiom_overshoots_where_appendf_does_not` runs the old expression and asserts the cursor lands *outside* the buffer and the next remainder is negative, then the same input through `spine_appendf()` and asserts it does not

That last one is the point: it fails if the defect is ever not a defect, and it documents why the other six exist.

44 of 44 pass. Clean `ubuntu:24.04` rebuild carries the same 4 warnings as develop.

### DIFF
--- a/spine.c
+++ b/spine.c
@@ -645,25 +645,25 @@ int main(int argc, char *argv[]) {
 
 	/* obtain the list of hosts to poll */
 	{
-		int remaining = MEGA_BUFSIZE - (qp - querybuf);
-		qp += snprintf(qp, remaining, "SELECT SQL_NO_CACHE id, device_threads, picount, picount/device_threads AS tppi FROM host AS h LEFT JOIN (SELECT host_id, COUNT(*) AS picount FROM poller_item GROUP BY host_id) AS pi ON h.id = pi.host_id");
+		size_t remaining = MEGA_BUFSIZE - (qp - querybuf);
+		spine_appendf(&qp, &remaining, "SELECT SQL_NO_CACHE id, device_threads, picount, picount/device_threads AS tppi FROM host AS h LEFT JOIN (SELECT host_id, COUNT(*) AS picount FROM poller_item GROUP BY host_id) AS pi ON h.id = pi.host_id");
 		remaining = MEGA_BUFSIZE - (qp - querybuf);
-		qp += snprintf(qp, remaining, " WHERE disabled = ''");
+		spine_appendf(&qp, &remaining, " WHERE disabled = ''");
 
 		remaining = MEGA_BUFSIZE - (qp - querybuf);
-		qp += snprintf(qp, remaining, " AND availability_method != %d", AVAIL_STREAM);
+		spine_appendf(&qp, &remaining, " AND availability_method != %d", AVAIL_STREAM);
 
 		if (!strlen(set.host_id_list)) {
 			qp += append_hostrange(qp, "h.id");	/* AND id BETWEEN a AND b */
 		} else {
 			remaining = MEGA_BUFSIZE - (qp - querybuf);
-			qp += snprintf(qp, remaining, " AND h.id IN(%s)", set.host_id_list);
+			spine_appendf(&qp, &remaining, " AND h.id IN(%s)", set.host_id_list);
 		}
 
 		remaining = MEGA_BUFSIZE - (qp - querybuf);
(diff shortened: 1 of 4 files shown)


somethingwithproof:

PR #595, opened 2026-09-03
End the MySQL thread on every poll_host exit

### LINKED ISSUE #594: poll_host() skips mysql_thread_end() when the device row is missing
`poll_host()` runs on a per-device thread and has three exit paths. Two call `mysql_thread_end()`, one does not.

```
poller.c:810   mysql_thread_end();     /* set.ping_only exit  */
poller.c:1962  mysql_thread_end();     /* normal exit         */
poller.c:523   return;                 /* device row missing: no thread_end */
```

The third is taken when the `host` lookup does not return exactly one row:

```c
if ((result = db_query(&mysql, LOCAL, q.query2)) != 0) {
	num_rows = mysql_num_rows(result);

	if (num_rows != 1) {
		db_free_result(result);
		/* releases both connections, frees six allocations */
		return;                      /* mysql_thread_end() is missing here */
	}
```

There is no `mysql_thread_init()` anywhere in the tree, so the client library initialises its thread-local state on first use. `mysql_thread_end()` is what releases it, and the MySQL C API requires it before a thread that used client functions exits. `poll_host()` has already run `db_query()` by the time it reaches this branch, so the state exists and is then abandoned.

## Reachability

`query2` is `SELECT ... FROM host WHERE id = %i AND deleted = ''`, so the branch fires whenever the device row is absent or soft-deleted between the poller picking up the device and the thread reading it. A device deleted mid-cycle does it, and so does a remote poller whose `host` table has not caught up with the main one.

Spine is thread-per-device, so this is once per affected device per polling cycle.

## Why it was missed

The same teardown is written out three times: six `SPINE_FREE` calls plus the two `db_release_connection` blocks, at lines 516, 789 and split across 1895/1974. The three copies are not in the same order and, as here, do not contain the same steps. Nothing reads the three side by side.

## Suggested fix
One release function used by all three exits, so they cannot diverge again.

somethingwithproof:
Closes #594. Stacked on #589; merge that first.

`poll_host()` leaves through three places. Two called `mysql_thread_end()`, one did not:

```
poller.c:810   mysql_thread_end();     /* set.ping_only exit */
poller.c:1962  mysql_thread_end();     /* normal exit        */
poller.c:523   return;                 /* device row missing */
```

There is no `mysql_thread_init()` in the tree, so the client library sets its thread-local state up on first use and `mysql_thread_end()` is what releases it. `poll_host()` has already issued `db_query()` before it reaches that branch, so the state exists and is then abandoned. Spine is thread-per-device, so it is once per affected device per cycle.

The branch fires when `SELECT ... FROM host WHERE id = %i AND deleted = ''` does not return exactly one row: a device deleted or soft-deleted between the poller picking it up and the thread reading it, or a remote poller whose `host` table has not caught up.

## Why it happened

The teardown was written out three times, six `SPINE_FREE` calls plus the same fourteen-line connection release. The three copies were not in the same order and, as here, did not contain the same steps. Nothing put them side by side.

## The change

`poll_host_release()` for the two early exits, taking pointers so the caller's pointers are nulled rather than left dangling.

The normal exit could not use it: it calls `buffer_output_errors(error_string, ...)` after the point where the helper frees `error_string`. Rather than reorder that, it shares `poll_host_release_connections()` only and keeps the rest of its sequence. So the fourteen-line connection block is now in one place across all three, and the six frees are shared across the two that can share them.

`poll_host()` drops from 1,620 to 1,570 lines.

## Verification

This one is not unit-testable: `mysql_thread_end()` is a libmysql call with no observable return, and reaching the branch needs a device row to vanish between two queries. Saying so rather than dressing up a weaker test as proof.

What was checked: every `return` in `poll_host()`, and its fall-off-the-end exit, now reaches `mysql_thread_end()`. Clean `ubuntu:24.04` build matching `ci.yml`, same four warnings as develop, `make check` green on all four binaries, 49 of 49 tests.

### DIFF
--- a/poller.c
+++ b/poller.c
@@ -222,6 +222,8 @@ void poll_host(int device_counter, int host_id, int host_thread, int host_thread
 	char temp_poll_result[BUFSIZE];
 	char temp_arg1[BUFSIZE];
 	char limits[SMALL_BUFSIZE];
+	char item_scope[SMALL_BUFSIZE];
+	char owner_scope[SMALL_BUFSIZE];
 
 	int  last_snmp_version = 0;
 	int  last_snmp_port    = 0;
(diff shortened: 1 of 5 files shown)


somethingwithproof:

PR #597, opened 2026-09-03
Spine correctness, hardening and poll_host decomposition

### LINKED ISSUE #564: Unchecked calloc() results in spine.c dereferenced immediately after allocation

somethingwithproof:
Consolidates ten pull requests into one review. Closes #564, #565, #566, #567, #581, #582, #583, #587, #588, #590, #594.

They were opened separately and every pair of them conflicted on `tests/unit/test_linked.c`, because each appended to the same registration array. Whichever merged first would have forced a rebase on the other nine. One review, one merge, one approval.

Nothing is dropped and nothing is squashed: **20 commits, one logical change each**, so the branch stays reviewable commit by commit.

## Fixes

**#581 — close-on-exec and bounded reap.** PR #542 removed both while adding unrelated poller fixes; its branch predated #557 and the base looks folded in rather than rebased. An inherited pipe write end stops a script's reader seeing EOF, so the thread blocks to `script_timeout` for a device that answered. The unbounded `waitpid` has no timeout at all, and the thread holds its `available_scripts` token while it waits.

**#564, #565, #566 — allocation checks, a log newline overflow, leaked result sets.** Four `calloc()` results dereferenced without a check; `spine_log()` putting a newline's terminator one byte past a `LOGSIZE` buffer; four settings helpers returning without `db_free_result()` on the NULL-row branch.

**#567 — `get_date_format()`** rebuilt an invariant string, with a malloc/free pair, on every log line. Built once at config load instead.

**#583 — results of 1024 bytes or more silently truncated to 1022.** `db_escape()` sized its destination from the wrong buffer.

**#582 — SNMPv3 authNoPriv and noAuthNoPriv non-functional since #373.** Cacti stores the literal `[None]` for an unselected protocol; treating that as an error made two of the three security levels unusable. A device that polls under cmd.php could not poll under spine.

**#588 — the `p += snprintf` idiom.** `snprintf` returns the length it *would* have written, so the first truncation moves the cursor past the buffer and the next remainder underflows to a huge `size_t`. 53 sites. `util.c:829-851` was already wrong in the stronger sense: it passed a fixed `BUFSIZE` to every call, so the bound never shrank.

**#590 — remote pollers ignored `dbonupdate`** and emitted the `VALUES()` upsert form MySQL deprecated in 8.0.20.

**#594 — `poll_host()` skipped `mysql_thread_end()`** when the device row was missing, leaking the client library's thread-local state once per affected device per cycle.

## Refactors

**`poller_item_from_row()`** — 62 lines of column mapping in the middle of the result loop. Twenty-one optional columns, each with a default a NULL must not overwrite: a row missing `snmp_port` has to leave 161 in place rather than zero it. None of that was reachable.

**#587 — `poll_host()` built the same six queries twice**, 141 and 142 lines differing by 33. Collapsed onto two scope helpers, then extracted as `poll_host_build_queries()`. The reindex assert was written out three times, once per operator, each with 24 identical lines around a one-line comparison; it is now `reindex_assert_failed()`.

`poll_host()` goes from **1,927 lines to 1,455**, and from 840 executable lines under gcov to 790. That is a first cut, not a decomposition: the two blocks left inside it are still 697 and 308 lines.

**#590 and #594 were both found by doing this.** Each was a case of duplicated code that had quietly stopped agreeing with its other copy. That is the argument for the refactor, and it is why they are in the same PR rather than filed as unrelated fixes.

## CI

A composite action for the build setup, and the unit tests run under ASan and UBSan.

## Test infrastructure

`tests/integration` and `tests/regression` were in the tree and in no `TESTS` list. `test_child_process_safety.sh` had been **failing on develop** since #542 removed the close-on-exec code it guards. It found the regression at the time and nobody saw it, because nothing ran it.

All seven are in `TESTS` now. Each exits 77 where its fixture is missing, which automake reports as SKIP, so `make check` is meaningful on a machine without docker and complete in CI with it. In-tree: 14 targets, 10 pass, 4 skip.

Two build modes were missing:

- `--enable-tsan`, because spine runs a thread per device over shared counters and a process-wide effective uid, and a data race here is the failure mode that costs the most to find any other way. It cannot be combined with `--enable-sanitizers`.
- `--enable-coverage`, so the gcov flags live in one place, with `make coverage` to read the per-function numbers back.

All three of `--enable-coverage`, `--enable-sanitizers` and `--enable-tsan` build and pass.

Out-of-tree builds were broken by two relative paths: the regression script and the golden fixture lookup both assumed the source directory. `make distcheck` now passes. The golden test was also skipping silently when it could not find its fixture, which meant it passed under distcheck **without running**; it fails rather than skips now.

## Tests

`test_linked` goes from 37 to **96**, across 7 binaries and 14 test targets, all against the shipped objects rather than copies of the logic. Coverage of what was extracted, measured with `gcov`:

```
poller_item_from_row             100.00% of 54
poll_host_build_queries          100.00% of 29
reindex_assert_failed            100.00% of 14
poll_host_release                100.00% of 10
poll_host_release_connections    100.00% of 9
poller_item_scope                100.00% of 6
poller_owner_scope               100.00% of 6
spine_appendf                     87.50% of 16
db_escape                        100.00% of 11
```

The teardown pair reached 100% by interposing on its two sinks. `db_release_connection()` and `mysql_thread_end()` return nothing to assert on, so a separate test binary wraps both with `ld --wrap` and records what was called:

```makefile
tests_unit_test_poll_host_release_LDFLAGS = \
    -Wl,--wrap=db_release_connection -Wl,--wrap=mysql_thread_end
```

That is worth noting beyond these two functions: most of what makes `poll_host()` unreachable is I/O with no observable result, and this is the cheaper way at it than extracting first.

`spine_appendf` sits at 87.5%. The two uncovered lines are the branch where `vsnprintf` returns negative on an encoding error. Reaching it portably is not possible without a libc-specific trick that would be fragile on another platform, so it is left uncovered and named rather than papered over.

Several tests fail against the old code on purpose, which is the point of them:

- reverting the `fcntl()` calls fails `test_cloexec_is_set_on_both_pipe_ends` and `test_pipe_is_not_inherited_across_exec`, the second reporting `0 != 0x1`, which is the child finding a descriptor it should not have
- `test_child_process_safety.sh` still fails against develop's `nft_popen.c`, checked by running it against those files directly
- `test_old_idiom_overshoots_where_appendf_does_not` runs the old expression and asserts the cursor lands outside the buffer
- `tests/golden/poll_host_queries.golden` is diffed line by line by `test_build_queries_matches_the_golden_capture`, so any change to the generated SQL fails the suite with the first differing line

Two tests were removed: `get_date_format()` no longer returns storage the caller owns, so the tests asserting that it did are gone with the behaviour.

## Verification

Clean `ubuntu:24.04` matching `ci.yml`: builds with the same four warnings as develop, `make check` green on all six binaries, 90 of 90 tests.

One warning changed text without changing count. `poller.c` reports `up to 4096 bytes` where develop says `up to 2047`, because the #583 fix correctly doubled `escaped_result` to hold a fully escaped value. Worth a follow-up: `result_string` is sized `RESULTS_BUFFER + SMALL_BUFSIZE`, which a fully escaped result can now exceed. That is a pre-existing risk this fix makes visible, not a new one, and it is still an improvement on truncating at 1022.

## Not included

- **#574** targets `1.2.x`. Different base, cannot combine.
- **#555** has review comments outstanding and stays on its own.


somethingwithproof:
Merge coordination for the current open set:

1. Merge #553.
2. Merge #555.
3. Rebase this PR onto the resulting `develop` and resolve the `poller.c`,
   `snmp.c`, `ping.c`, `util.c`, and `configure.ac` overlap once.
4. Fold the child-registry ownership work tracked by #609 into this PR after
   the active `php.c` work is complete.

#574 targets `1.2.x` and proceeds independently. None of the four open PRs is
obsolete, so none should be closed. Use a rebase, not a merge-style branch
update.

somethingwithproof:
Code-scanning overlap check: all 32 currently open alerts are `cpp/overflowing-snprintf` findings in `util.c` (lines 1045–1192 on `develop`). This PR’s #588 change replaces those executable `p += snprintf(...)` sites with `spine_appendf`, so I am not opening a duplicate security PR. After this branch is rebased and CodeQL reruns, we should verify that alerts 2502–2533 close automatically.

xmacan (approved):

somethingwithproof:
Reorganization status: the reviewable pieces have been extracted into focused PRs with independent tests and rollback boundaries.

Ready/open:
- subprocess ownership: #614
- source layout foundation: #617
- shared CI, sanitizers, and Linux capability coverage: #618
- ICMP/resolver cleanup and retry safety: #619
- SNMPv3 session construction and NULL multi-get safety: #620
- core allocation, logging, buffer, remote-push, and hex-result safety: #621

All Copilot inline threads across #597, #614, #617, #618, #619, #620, and #621 are resolved. Fresh Copilot rereviews have been requested after the latest fixes on #618, #619, #620, and #621.

Remaining dependency-gated slices:
- SQL result/escaping safety is rebased locally onto the core-safety series and should be published after #621 merges, avoiding duplicate core commits.
- child runtime cleanup follows #614 because both change subprocess ownership semantics.
- poll_host decomposition follows the safety slices and #617 so it can be rebased as a structural-only change.

This PR remains the umbrella/accounting record; it should not be merged as a second copy of the extracted changes.

somethingwithproof:
Superseded by narrower, independently reviewable PRs to reduce review bandwidth:

- build and CI infrastructure: #618
- ICMP correctness and safety: #619
- SNMP correctness: #620
- subprocess ownership and child runtime: #614

The remaining core, SQL, and poll-host work is preserved on the existing topic branches and deliberately deferred rather than dropped. Closing this umbrella removes duplicate review surface; its branch and commits remain recoverable.

